PR #3, opened 2016-10-11
10.0 Port all intrastat modules

alexis-via:
This PR will contain the port of all the modules of the project to Odoo v10.

For the moment, only intrastat_base has been ported.


alexis-via:
I ported product_harmonized_system, but I have a strange warning when loading the module:

```
2016-10-11 16:11:48,758 6885 WARNING test_01 odoo.fields: Field hs.code.display_name depends on itself; please fix its decorator @api.depends().
```

If you have a clue...

hbrunn:
@alexis-via I think this is caused by https://github.com/OCA/intrastat/pull/3/files#diff-b0f1105b01cd5c85d3ad003c0a061e16R15 - the ORM uses this as @depends for its own `_compute_display_name` in https://github.com/OCA/OCB/blob/10.0/odoo/models.py#L1511.
You're better off in my opinion with not touching `rec_name` and `_compute_display_name` at all, just override `name_get` to do what you need there, and override the computation function only to set the proper depends

pedrobaeza:
Yeah, that's something that happens in v10 with the special field `display_name`. Look here for a same issue detected by an Odoo employee that has worked on some community modules migration for Hacktoberfest: https://github.com/OCA/social/pull/108#discussion_r83491956

alexis-via:
@hbrunn Thank you very much for your advice. I thought that name_get() was kind of deprecated and it was bettter to use a computed 'display_name' field, cf https://www.odoo.com/fr_FR/forum/how-to/developers-13/how-to-override-name-get-method-in-new-api-61228
But maybe that info was wrong...

hbrunn:
@alexis-via travis errors van be fixed by including an update for `.travis.cfg`: https://github.com/OCA/intrastat/blob/10.0/.travis.yml#L35

hbrunn:
@alexis-via how do you see this? Should we merge what we have already? I'd be in favor of this, because then other people can give you a hand, while if we keep one PR to rule them all, you're alone.

alexis-via:
@hbrunn yes, good idea !

hbrunn:
okay, then I'll merge this as it is without squashing, this way, you can simply keep your branch and create a new PR when there are new commits

PR #7, opened 2016-11-18
9.0 mig product harmonized system

thomaspaulb:
Migration of this module and addition of tests.

### DIFF
--- a/product_harmonized_system/tests/__init__.py
+++ b/product_harmonized_system/tests/__init__.py
@@ -1,0 +1,1 @@
+from . import test_hs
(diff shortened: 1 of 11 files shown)


product_harmonized_system/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import test_hs
```
pedrobaeza: Line at the end

thomaspaulb: Added an encoding line at the beginning, is that what you mean?

hbrunn (approved):

pedrobaeza (approved):
Squashing commits on merge.

PR #9, opened 2016-12-02
[ADD] intrastat_base tests

thomaspaulb:
Added tests.

### DIFF
--- a/intrastat_base/tests/test_all.py
+++ b/intrastat_base/tests/test_all.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8 -*-
+
+from openerp.tests.common import TransactionCase
+
+
+class TestIntrastatBase(TransactionCase):
+    """Tests for this module"""
+
+    def setUp(self):
+        super(TestIntrastatBase, self).setUp()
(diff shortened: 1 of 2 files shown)


intrastat_base/tests/test_all.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+
+from openerp.tests.common import TransactionCase
+
+
+class TestIntrastatBase(TransactionCase):
+    """Tests for this module"""
+
+    def setUp(self):
+        super(TestIntrastatBase, self).setUp()
```
hbrunn: this is a noop

hbrunn (approved):
only a small nag

pedrobaeza:
Why tests are not in their respective modules?

astirpe (approved):
The setUp() method should be removed. Other than that LGTM.

thomaspaulb:
@astirpe @hbrunn Thanks and done.
@pedrobaeza Because the functionality tested is in this module, not in others. It's just to increase coverage.

PR #14, opened 2017-07-24
[8.0] FIX intrastat transaction on refund created from an invoice via the wizard

alexis-via:
Go to the menu "Accounting > Customers > Customer Invoice" and create a customer invoice: the field "Intrastat Transaction Type" gets the default value for Customer Invoices.

Valide the customer invoice.

Click on the button "Refund invoice" and select refund method = "Create a draft refund"

The value of the field "Intrastat Transaction Type" on the refund is the same as the value on the customer invoice -> This is a bug !

Why this bug ? Because, when the customer refund is created via the wizard, the value of the context is still {'type': 'out_invoice'}, so it gets the default value for customer invoices.

2 possible solutions:
1) inherit the method _prepare_refund() to set the value of intrastat_transaction_id to the default value for refunds
2) don't put any default value for the field intrastat_transaction_id because, when we generate the intrastat declaration, it will read the default value from the company configuration. With this solution, we would say to the users: only set this field on an invoice when the default configuration for the company is not appropriate.

This PR implements solution 2).

@luc-demeyer Are you ok with this ? If yes, I'll up-port it to v10.

### DIFF
--- a/intrastat_product/models/account_invoice.py
+++ b/intrastat_product/models/account_invoice.py
@@ -34,7 +34,6 @@ class AccountInvoice(models.Model):
              "commercial terms used in international transactions.")
     intrastat_transaction_id = fields.Many2one(
         'intrastat.transaction', string='Intrastat Transaction Type',
-        default=lambda self: self._default_intrastat_transaction_id(),
         ondelete='restrict',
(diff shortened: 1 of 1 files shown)


luc-demeyer:
👍

PR #15, opened 2017-07-25
[FIX] Update hs_code to hs_code_id

leemannd:
Hello,
In [odoo/delivery](https://github.com/odoo/odoo/blob/10.0/addons/delivery/views/product_template_view.xml#L10) you can find the hs_code field. It is also introduced by https://github.com/OCA/intrastat/blame/10.0/product_harmonized_system/views/product_template.xml#L17 in the product.template view.

The goal of this PR is to use the same name for this field in OCA and Odoo's base code source.

### DIFF
--- a/product_harmonized_system/README.rst
+++ b/product_harmonized_system/README.rst
@@ -51,6 +51,7 @@ Contributors
 
 * Alexis de Lattre, Akretion <alexis.delattre@akretion.com>
 * Luc De Meyer, Noviat <info@noviat.com>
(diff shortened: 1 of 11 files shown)


leemannd:
@alexis-via @luc-demeyer If I can have your review?

gurneyalex (changes requested):
This cannot work : in delivery the field hs_code = fields.Char

You cannot change the field type like you do. IMO, you should

* change the field hs_code to be a related field (readonly) to use the hs code name from this module
* change the form to hide the hs_code field

PR #16, opened 2017-07-25
region attrs invisible

luc-demeyer:
some localisation modules (e.g. Belgium) require the region field.
Other localisations also require the region concept for the intrastat declaration but this requirement is fulfilled via other means, e.g.
- France : departments
- Spain : states

This creates a confusing User Interfaces with region, department and state fields on several objects when running in a multi-company setup with multiple intrastat localisations.

This PR solves this by implementing the display/non-display logic of the country field to the 'intrastat_product' module whereby a list of country codes in the "attrs" attribute determines the visibility.

### DIFF
--- a/intrastat_product/models/account_invoice.py
+++ b/intrastat_product/models/account_invoice.py
@@ -43,6 +43,8 @@ class AccountInvoice(models.Model):
     src_dest_country_id = fields.Many2one(
         'res.country', string='Origin/Destination Country',
         ondelete='restrict')
+    src_dest_country_code = fields.Char(
+        related='company_id.country_id.code')
     src_dest_region_id = fields.Many2one(
         'intrastat.region', string='Origin/Destination Region',
         default=lambda self: self._default_src_dest_region_id(),
--- a/intrastat_product/models/account_invoice.py
+++ b/intrastat_product/models/account_invoice.py
@@ -1,28 +1,9 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    Intrastat Product module for Odoo
-#    Copyright (C) 2011-2016 Akretion (http://www.akretion.com)
-#    Copyright (C) 2009-2016 Noviat (http://www.noviat.com)
-#    @author Alexis de Lattre <alexis.delattre@akretion.com>
-#    @author Luc de Meyer <info@noviat.com>
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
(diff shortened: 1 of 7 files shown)


pedrobaeza:
Please change version number

intrastat_product/models/account_invoice.py
```diff
@@ -43,6 +43,8 @@ class AccountInvoice(models.Model):
     src_dest_country_id = fields.Many2one(
         'res.country', string='Origin/Destination Country',
         ondelete='restrict')
+    src_dest_country_code = fields.Char(
+        related='company_id.country_id.code')
```
pedrobaeza: Please add `readonly=True` for avoiding undesired write back.

intrastat_product/models/res_company.py
```diff
@@ -46,6 +46,8 @@ class ResCompany(models.Model):
     intrastat = fields.Char(
         string='Intrastat Declaration', store=True, readonly=True,
         compute='_compute_intrastat')
+    country_code = fields.Char(
+        related='country_id.code')
```
pedrobaeza: `readonly=True`

intrastat_product/models/stock_warehouse.py
```diff
@@ -29,6 +29,8 @@ class StockWarehouse(models.Model):
     region_id = fields.Many2one(
         'intrastat.region',
         string='Intrastat region')
+    country_code = fields.Char(
+        related='company_id.country_id.code')
```
pedrobaeza: `readonly=True`

pedrobaeza (changes requested):

pedrobaeza (approved):

intrastat_product/views/account_invoice.xml
```diff
@@ -18,7 +18,9 @@
                  attrs="{'invisible': [('intrastat', '!=', 'extended')]}"
                  widget="selection"/>
           <field name="src_dest_country_id" string="Destination Country"/>
-          <field name="src_dest_region_id" string="Origin Region" invisible="1"/>
+          <field name="src_dest_country_code" invisible="1"/>
+          <field name="src_dest_region_id" string="Origin/Destination Region"
+                 attrs="{'invisible': [('src_dest_country_code', 'not in', ['BE'])]}"/>
```
alexis-via: I don't agree on that one.
I am a belgian company and I ship from Belgium to France. On the customer invoice:
src_dest_country_code = FR
src_dest_region_id = the Belgian region from where the goods are shipped
-> so it should not be invisible here !

intrastat_product/views/account_invoice.xml
```diff
@@ -42,7 +44,9 @@
                  attrs="{'invisible': [('intrastat', '!=', 'extended')]}"
                  widget="selection"/>
           <field name="src_dest_country_id" string="Origin/Destination Country"/>
-          <field name="src_dest_region_id" string="Origin/Destination Region" invisible="1"/>
+          <field name="src_dest_country_code" invisible="1"/>
+          <field name="src_dest_region_id" string="Origin/Destination Region"
+                 attrs="{'invisible': [('src_dest_country_code', 'not in', ['BE'])]}"/>
         </xpath>
```
alexis-via: I don't agree either, for the same reason.

alexis-via (changes requested):

luc-demeyer:
@alexis-via 
I understand the confusion. 
Technically the PR is ok since src_dest_country_code is related to the company country but on the invoice the name could suggest that it is related to the src_dest_country_id field.
I'll adapt the PR with  a field name that is less confusing.

alexis-via (approved):

PR #20, opened 2017-07-27
[10.0]Add hs_code_link

leemannd:
Odoo added a field char field `hs_code` in the delivery module.
It results in a duplicate field from `product_harmonized_system` and it can be
seen twice in the `product.template` form view.

The goal of this module is to hide Odoo's field in the form and make it related
to the OCA's one if it is used elsewhere.

This module is set as auto_install if the following modules are installed:

* product_harmonized_system
* delivery

### DIFF
--- a/hs_code_link/models/product.py
+++ b/hs_code_link/models/product.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, fields
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    hs_code = fields.Char(
+        related='hs_code_id.hs_code',
+        readonly=True,
+        store=True,
+    )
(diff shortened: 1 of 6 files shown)


leemannd:
@alexis-via @luc-demeyer @gurneyalex Can you review please?

hs_code_link/models/product.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import models, fields
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    hs_code = fields.Char(
+        related='hs_code_id.hs_code',
+        readonly=True,
```
gurneyalex, line 13: I think we want store=True too  in case this is used in some reports.

hs_code_link/views/product_views.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+  <record id="view_template_hide_hs" model="ir.ui.view">
+    <field name="name">product.template.hs.hiden</field>
```
gurneyalex: nitpicking: "hidden"

gurneyalex (changes requested):

gurneyalex (approved):

guewen (approved):
Thanks!

tafaRU (approved):

PR #22, opened 2017-09-20
[FIX] intrastat_product: state can be None

cubells:
Company's country field is not required.
If country is not set, an error is raised.

cc @Tecnativa


pedrobaeza:
Change module version

leemannd (approved):
LGTM Thanks!

pedrobaeza (approved):

PR #24, opened 2017-10-07
[8.0] skip invoice lines with value 0

luc-demeyer:


cubells (approved):
:+1: 

LGTM

Code review and test.

luc-demeyer:
@cubells
I changed the check on 'price_unit' with value 0.0 into 'price_subtotal'.
By doing so we also cover the case of invoice lines with 100% discount.
Let me know if it's ok so that we can merge (and forward port to v10)

intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -501,6 +501,9 @@ def _gather_invoices(self):
             total_inv_weight = 0.0
             for inv_line in invoice.invoice_line:
 
+                if not inv_line.price_subtotal:
+                    continue
+
```
cubells: Remove all the white spaces of this line to avoid lint errors.

cubells (approved):
@luc-demeyer 

If you remove white spaces all is ok for me.

intrastat_product/models/intrastat_product_declaration.py
pedrobaeza: A better expression is:
```python
for inv_line in invoice.invoice_line.filtered('price_subtotal'):
    ...
```

pedrobaeza:
Travis is red

PR #25, opened 2017-12-27
res_company: fix intrastat_dispatches selection function

CGenie:


rvalyi (approved):

pedrobaeza (changes requested):
As this is not impacting right now in the functionality, I fast-track without changing module version number

pedrobaeza (approved):

Issue #23, opened 2017-10-03
Migration to version 11.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

# Modules to migrate

- [ ] hs_code_link
- [x] intrastat_base - ~~By @BT-kaberer - #28~~ - By @rvalyi - #37  
- [x] intrastat_product - ~~By @BT-kaberer - #30~~ By @rvalyi - #47 
- [x] product_harmonized_system - ~~By @BT-kaberer - #29~~ - By @rvalyi - #37

ghost:
Hey @pedrobaeza 

I will start migrating 
 intrastat_base
 intrastat_product
 product_harmonized_system
to version 11.0

pedrobaeza:
OK, noted

emagdalenaC2i:
@pedrobaeza intrastat_product - By @BT-kaberer - #30 is closed and superseded by #47

product_harmonized_system - By @BT-kaberer - #29 is closed and superseded by #47 that is already merged

luc-demeyer:
closing since 11.0 migration has been completed a while ago.

PR #28, opened 2018-03-09
11.0 mig intrastat base

ghost:
#23


feketemihai:
@BT-kaberer Why we are keeping the incompatibility between report_intrastat from main branch and this, just that this module adds the same field in countries and data, doesn't mean that we should double that, and not to depends on that module...

ghost:
@feketemihai 
intrastat_base act as abstract module for localization. it not only adds a new field and data in res.country. 
it defines an abstract report model that is used in the localization modules. l10n_fr_intrastat_service,, l10n_fr_intrastat_product, l10n_be_intrastat_product: 

now my assumption is that if it would depend on standard instrastat_report there would be confusion between two different intrastat reporting from the original base module and the localization. Because both modules would be installed. And that's why it's marked as conflicting.
https://github.com/odoo/odoo/blob/11.0/addons/report_intrastat/report/report_intrastat_report.py
vs.
https://github.com/OCA/intrastat/blob/10.0/intrastat_base/models/intrastat_common.py

maybe @alexis-via can shed more light to this

intrastat_base/README.rst
```diff
@@ -0,0 +1,80 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
astirpe, line 1: Prefer .png images, see https://raw.githubusercontent.com/OCA/maintainer-tools/52f41366ab39ea90e66f3e3a2b5cd7691b4219df/template/module/README.rst

intrastat_base/README.rst
```diff
@@ -0,0 +1,80 @@
...
+With this module, the country field on partners becomes a required field.
+
+It adds an option *Exclude invoice line from intrastat if this tax is present*
+on taxes.
+
+It adds a tab *Intrastat* on the company form view.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/227/10.0
```
astirpe, line 46: 11.0

intrastat_base/__init__.py
```diff
@@ -0,0 +1,2 @@
+# -*- coding: utf-8 -*-
```
astirpe, line 1: Remove coding utf-8 line, since it's already the standard for Python3

intrastat_base/__manifest__.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
```
astirpe, line 1: Remove coding utf-8 line, since it's already the standard for Python3

intrastat_base/models/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
```
astirpe, line 1: Remove coding utf-8 line, since it's already the standard for Python3

intrastat_base/models/intrastat_common.py
```diff
@@ -0,0 +1,168 @@
+# -*- coding: utf-8 -*-
```
astirpe, line 1: Remove coding utf-8 line, since it's already the standard for Python3

intrastat_base/models/res_country.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
```
astirpe, line 1: Remove coding utf-8 line, since it's already the standard for Python3

intrastat_base/models/res_config_settings.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
```
astirpe, line 1: Remove coding utf-8 line, since it's already the standard for Python3

intrastat_base/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Intrastat Reporting Base',
+    'version': '11.0.1.0.0',
+    'category': 'Intrastat',
+    'license': 'AGPL-3',
+    'summary': 'Base module for Intrastat reporting',
+    'author': 'Akretion,Odoo Community Association (OCA), brain-tec AG',
+    'website': 'http://www.akretion.com',
```
astirpe, line 13: Change the website to `https://github.com/OCA/intrastat`

astirpe (changes requested):

astirpe:
Also it would be nice to have some tests (for example the tests proposed in https://github.com/OCA/intrastat/pull/9) included in this module.

rvalyi:
closing this one as it is superseded by https://github.com/OCA/intrastat/pull/37 which takes the last review into account...

PR #31, opened 2018-04-10
[10.0]FIX account.config.settings

luc-demeyer:
Fix invalid domain in account.config.settings view

### DIFF
--- a/intrastat_product/views/account_config_settings.xml
+++ b/intrastat_product/views/account_config_settings.xml
@@ -14,6 +14,7 @@
           <field name="intrastat_transaction_out_refund"/>
           <field name="intrastat_transaction_in_invoice"/>
           <field name="intrastat_transaction_in_refund"/>
+          <field name="country_id" invisible="1"/>
           <field name="intrastat_region_id" domain="[('country_id','=', country_id)]" invisible="1"/>
           <field name="intrastat_transport_id"
                  attrs="{'required': [('intrastat', '=', 'extended')], 'invisible': [('intrastat', '!=', 'extended')]}"/>
--- a/intrastat_product/__manifest__.py
+++ b/intrastat_product/__manifest__.py
@@ -6,7 +6,7 @@
 
 {
     'name': 'Intrastat Product',
-    'version': '10.0.1.1.0',
+    'version': '10.0.1.2.0',
     'category': 'Intrastat',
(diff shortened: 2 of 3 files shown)


pedrobaeza (approved):

intrastat_product/views/account_config_settings.xml
```diff
@@ -14,6 +14,7 @@
           <field name="intrastat_transaction_out_refund"/>
           <field name="intrastat_transaction_in_invoice"/>
           <field name="intrastat_transaction_in_refund"/>
+          <field name="country_id" invisible="1"/>
```
moylop260, line 17: I have a doubt.

This field needs be added in order to fix a domain error for `intrastat_region_id` field.
But I don't understand what is the reason to adds a field with domain but invisible in the case of `intrastat_region_id`

What about remove from the view `intrastat_region_id` since that is not used from other domain or another field?
And both spent time to compute them the related (Because they are not using `store=True`)

luc-demeyer:
The intention of the 'intrastat_product' module is to have a common base for the localization modules.
The 'region' of destination/origin is required for the intrastat declarations of 30% of the EU member states. Because it's not mandatory for all states we decided when creating the intrastat_product module a couple of years ago to add this field to the view but as invisible so that a localisation module can simply turn it on. (cf. https://github.com/luc-demeyer/noviat-apps/blob/10.0/l10n_be_intrastat_product/views/account_config_settings.xml for an example)

But I agree that the approach is not optimal, especially in a multi-country system where this field will become visible for all countries once it's turned one by e.g. the belgian localization (and hence the french and spanish users will be confused since in France the localisation module does not use the 'region' field but the 'department' field whereas in Spain the standard 'state' field is used for this purpose.)

As an alternative we could change the view in the intrastat_product module as follows:

          <field name="intrastat_region_id" domain="[('country_id','=', country_id)]"
                 attrs="{'invisible': [('country_code', 'not in', ['BE'])]}"/>

The consequence of this approach is that a editors of a localization module using the region field should make PR to add their country code to the list.

Any feedback on the preferred way forward is welcomed so that the Intrastat PSC can use this input to make a decision on the best approach to fix this issue.

alexis-via (approved):

alexis-via:
@luc-demeyer I prefer the current "simple" method with just visible/invisible.

PR #35, opened 2018-07-06
[10.0][FIX] Several very important fixes

alexis-via:
Here is the scenario of the bug:
1) create a customer invoice with "Intrastat Transaction Type" = False and validate it
2) click on the "Refund" button and follow the wizard to create a draft refund
3) In the draft refund created, "Intrastat Transaction Type" has the default value for customer invoices, instead of having the default value for customer refunds

Code analysis: when you look at the code of the refund wizard, you see that it doesn't set the 'type' in the context, so the value of 'type' in the context is still 'out_invoice', so the default value for "Intrastat Transaction Type" is the one of Customer Invoices.

For me, the solution is just to not put any default value for the field "Intrastat Transaction Type". The code that generate the transaction lines of the intrastat declaration will use the default value for invoices where the field ""Intrastat Transaction Type" is empty.

### DIFF
--- a/intrastat_product/data/intrastat_transport_mode.xml
+++ b/intrastat_product/data/intrastat_transport_mode.xml
@@ -30,7 +30,6 @@
       <field name="code">7</field>
       <field name="name">Fixed installations</field>
       <field name="description">Fixed transport installations (e.g. pipelines, high-tension cables)</field>
-      <field name="description"></field>
     </record>
     <record id="intrastat_transport_8" model="intrastat.transport_mode">
       <field name="code">8</field>
--- a/intrastat_product/models/account_invoice.py
+++ b/intrastat_product/models/account_invoice.py
@@ -22,12 +22,12 @@ class AccountInvoice(models.Model):
     src_dest_country_id = fields.Many2one(
         'res.country', string='Origin/Destination Country',
         compute='_compute_intrastat_country',
-        store=True,
+        store=True, compute_sudo=True,
         help="Destination country for dispatches. Origin country for "
         "arrivals.")
     intrastat_country = fields.Boolean(
         compute='_compute_intrastat_country',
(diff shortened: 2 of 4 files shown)


pedrobaeza:
I think it's better to overload the method that creates the refund and force the value, don't you think?

alexis-via:
If we do that, the problem will happen everytime the 'type' is not properly set in the context, which may happen at several other places.
@luc-demeyer What's your point of view on this ?

alexis-via:
I added compute_sudo=True for the fields that use information from partners and store it on invoices, to fix multi-company problems.

alexis-via:
@luc-demeyer I found one more very important bug: the computation of region was wrong on arrival intrastat declaration: we were taking source location instead of destination location on the move. I'm very surprised we didn't find it earlier ; it seems it was already there on v8. It's one of my customers who found it.

alexis-via:
I backported the fix on region to v8 here: https://github.com/OCA/intrastat/pull/36
In fact, I was not impacted by the bug on v8 because I don't use _get_region in l10n_fr_intrastat_product, which explains why my users were not affected.

luc-demeyer:
I have performed code review and tested this PR.
During the test I found a bug that is unrelated to this PR but since the fix is straightforward I propose to merge it into this PR (cf. https://github.com/akretion/intrastat/pull/1) since this PR already combines a couple of fixes.
After merging https://github.com/akretion/intrastat/pull/1 into this PR, this PR can be merged.

luc-demeyer (approved):

alexis-via:
@pedrobaeza Could you add write access to @luc-demeyer on this repo ? Github says "At least 1 approving review is required by reviewers with write access" and luc approved, but he is not counted.

alexis-via:
and by the way, I think it's not fair to count only people with write access, because there are too few of them and this project has a very small audience.

PR #38, opened 2018-07-30
11.0 mig product harmonized system

rvalyi:
supersedes https://github.com/OCA/intrastat/pull/29
additionally, it

- removes the Python 2 utf-8 headers
- fixes the README icon
- apply the latest v10.0 fixes from Noviat

cc @BT-kaberer @feketemihai @alexis-via @astirpe @luc-demeyer


product_harmonized_system/views/hs_code.xml
```diff
@@ -0,0 +1,72 @@
...
+
+    <!-- Action for H.S. code -->
+    <record id="hs_code_action" model="ir.actions.act_window">
+      <field name="name">H.S. Codes</field>
+      <field name="res_model">hs.code</field>
+      <field name="view_mode">tree,form</field>
+    </record>
+
+    <!-- Menu entry for H.S. code -->
+    <!-- TODO: find a way to put a menu entry without depending on another module ?
```
astirpe, line 67: An idea could be adding a button or a link in the `'res.config.settings'` that triggers the `hs_code_action`. This way you don't need to depend on another module.

product_harmonized_system/__manifest__.py
```diff
@@ -0,0 +1,30 @@
+# © 2018 brain-tec AG (http://www.braintec-group.com)
```
astirpe: Could you replace the symbol `©` with the string `Copyright` in the headers?

product_harmonized_system/models/hs_code.py
```diff
@@ -0,0 +1,74 @@
...
+    _order = "local_code"
+    _rec_name = "display_name"
+
+    hs_code = fields.Char(
+        string='H.S. Code', compute='_compute_hs_code', readonly=True,
+        help="Harmonized System code (6 digits). Full list is "
+        "available from the World Customs Organisation, see "
+        "http://www.wcoomd.org")
+    description = fields.Char(
+        'Description', translate=True,
```
astirpe: No need of `'Description'`

product_harmonized_system/models/hs_code.py
```diff
@@ -0,0 +1,74 @@
...
+        help="Short text description of the H.S. category")
+    display_name = fields.Char(
+        compute='_compute_display_name_field', string="Display Name",
+        store=True, readonly=True)
+    local_code = fields.Char(
+        string='Local Code', required=True,
```
astirpe: No need of `string='Local Code'`

product_harmonized_system/models/hs_code.py
```diff
@@ -0,0 +1,74 @@
...
+    hs_code = fields.Char(
+        string='H.S. Code', compute='_compute_hs_code', readonly=True,
+        help="Harmonized System code (6 digits). Full list is "
+        "available from the World Customs Organisation, see "
+        "http://www.wcoomd.org")
+    description = fields.Char(
+        'Description', translate=True,
+        help="Short text description of the H.S. category")
+    display_name = fields.Char(
+        compute='_compute_display_name_field', string="Display Name",
```
astirpe: No need of `string="Display Name"`

astirpe (changes requested):
Could you also add the readme files? See latest template https://github.com/OCA/maintainer-tools/tree/master/template/module/readme

product_harmonized_system/models/hs_code.py
luc-demeyer: Code review & tested. 
Remarks of @astirpe are valuable hence would be nice to include but imho not blocking for merging this PR.

luc-demeyer (approved):

product_harmonized_system/models/hs_code.py
astirpe: @luc-demeyer ... however it's stated in the OCA Guidelines: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#fields

luc-demeyer: @astirpe  I fully agree, hence if @rvalyi  is prepared to adjust the PR accordingly than it's aligned more closely to the OCA guidelines. Imho not blocking though for merge. I prefer to have this one merged asap so that we can focus on the larger work (port of intrastat_product for which I would first like to have a couple of major fixes and enhancements in the 10.0 version merged and use this as the basis for the port).

luc-demeyer:
@rvalyi @astirpe 
Merged whereby I also added the suggestions of andrea stirpe.
A remaining point is the addition of a menu entry where andrea suggests to work via a config option.
An alternative could be an auto-install module depending on 'sale' and 'product_harmonized_system'.
What do you think ?

PR #39, opened 2018-08-08
[10.0][FIX]reporting level

luc-demeyer:
The _prepare_grouped_fields method does not pass the declaration object. As a consequence the localisation methods cannot add logic based upon declaration fields to add extra data for the local declaration.

A concrete example:

In the Belgian l10n_be_intrastat_product module we need to add the incoterms to the declaration lines in case of an extended declaration (i.e. declaration.reporting_level == 'extended').
Since the declaration is not passed to the _prepare_grouped_fields method we cannot perform the check and hence the incoterms are not added.

This PR fixes this issue but the proposed fix requires also a change to the localisation modules in case the inherited module call one of the methods changed by this PR:
- _prepare_declaration_line
- _prepare_grouped_fields

Since the localisation modules are in other repos, a simultaneous commit requires manual attention (can be avoided with a less clean solution than this PR).

### DIFF
--- a/intrastat_product/__manifest__.py
+++ b/intrastat_product/__manifest__.py
@@ -6,7 +6,7 @@
 
 {
     'name': 'Intrastat Product',
-    'version': '10.0.1.2.1',
+    'version': '10.0.1.3.0',
     'category': 'Intrastat',
     'license': 'AGPL-3',
     'summary': 'Base module for Intrastat Product',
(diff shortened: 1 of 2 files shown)


luc-demeyer:
I have now reworked the fix so that the existing localization modules will not break.
The patch is important since extended declarations will not work without this patch.
I'll backport to 8.0 as soon as this PR is merged and we must ensure to have this patch included in the forward port to 11.0.

alexis-via (approved):

PR #40, opened 2018-08-08
add oca icon

luc-demeyer:


pedrobaeza:
Well, the icon is not really needed, but if we put one, maybe it's better to put an specific one. I can design it. Can we consider this a subset of Accounting for starting from that icon, or should it be an independent icon?

luc-demeyer:
we see the oca icon on the majority of the oca modules and imho it's nicer to hava an icon than nothing at all. If the use of the oca icon is no longer part of the practice for oca modules than we should indeed move to specific ones. Since the intrastat suite serves for european specific statistics  closely related with accounting (with cross controls to tax declarations in several countiries): what about an accouning logo with the european flag on it ?

pedrobaeza:
I have switched to an specific icon based on v12 account one.

luc-demeyer:
I don't think we should have an american dollar sign on intrastat which is a european declaration. I propose that we simply stay with the good old OCA icon. As an alternative we can use the EU flag.

pedrobaeza:
As said, that icon is the account one. It doesn't matter about the dollar sign. The generic OCA icon is the same for all, so I prefer none over that one.

luc-demeyer:
let's than close this PR.

PR #45, opened 2018-08-16
[8.0][FIX] get_hs_code_recursively broken on declaration

luc-demeyer:
The intrastat declaration is broken in case no hs code is present on invoice lines as a consequence of https://github.com/OCA/intrastat/commit/33a92dcc4becd1e13b64d029b1b2549a747bee5b.


rvalyi (approved):
approved for v8.
However it is important to notice that on v10, the code is still taken from the template: 
https://github.com/OCA/intrastat/blob/10.0/intrastat_product/models/account_invoice.py#L68
@luc-demeyer @alexis-via @sebastienbeau so what do you think, should we take it from the variant on v10 too?

luc-demeyer:
@rvaly 
v10 should be the same. I'll have a look and make a PR.
Can you also adapt your PR (https://github.com/OCA/intrastat/pull/38) for the 11.0 port and move the method to product.product ?

PR #50, opened 2018-10-05
[FIX] Avoid error "The record has been modified, your changes will be discarded"

alexis-via:
Avoid error "The record has been modified, your changes will be discarded. Are you sure you want to leave this page ?"
 
@luc-demeyer This is the bug you were looking for...

### DIFF
--- a/intrastat_product/views/intrastat_product_declaration.xml
+++ b/intrastat_product/views/intrastat_product_declaration.xml
@@ -32,7 +32,7 @@
             </div>
             <group name="top-block">
               <group name="properties-1">
-                <field name="year" widget="char"/>
+                <field name="year"/>
                 <field name="month"/>
                 <field name="type"/>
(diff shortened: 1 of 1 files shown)


rafaelbn (approved):
:+1

luc-demeyer:
@alexis-via 
I was just about to make the exact same PR but you took me in speed.
The widget="char" on an integer works ok in 8.0 but not in 10.0.
This PR brings us back to the 'ugly' view where the year 2018 is presented with a dot after the 2 -> 2.018.

We could work around it by turning this field into a real char field but we may break some localisation modules.

I propose that we take this approach in the 11.0 port.

@rvalyi 
Do you agree to modify https://github.com/OCA/intrastat/pull/47 like that ?

alexis-via:
I'm OK to modify this field to char in v11.

PR #55, opened 2018-12-04
[FIX] 10.0 make hs_code company_id optional

PierrickBrun:


product_harmonized_system/models/hs_code.py
```diff
@@ -32,7 +32,7 @@ class HSCode(models.Model):
         "has a few additional digits to extend the H.S. code.")
     active = fields.Boolean(default=True)
     company_id = fields.Many2one(
-        'res.company', string='Company', readonly=True, required=True,
+        'res.company', string='Company',
         default=lambda self: self.env['res.company']._company_default_get(
             'hs.code'))
```
bealdav: Please fix argument in default

https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/res/res_company.py#L290-L295

alexis-via (approved):

luc-demeyer (approved):

product_harmonized_system/models/hs_code.py
bealdav: 'hs.code' param is useless

PR #60, opened 2019-02-02
[8.0] add 'type' field to lines views

luc-demeyer:
The belgian 2019 declaration requires different fields for arrivals and dispatches declarations.
As a consequences it is required to have the 'type' field available in the lines views so that it can be used in the attrs.


luc-demeyer:
@alexis-via 
Merged since urgent PR to get 2019 belgian intrastat declaration working.
I tested this PR with l10n_fr_intrastat_product to ensure that the french localisation is not broken.

PR #65, opened 2019-03-13
12.0 mig intrastat base

rvalyi:
standard v12 migration


alexis-via (approved):

alexis-via:
@rvalyi I don't know what you did, but this PR doesn't include code that is present in origin/11.0, for example in intrastat_base, it doesn't include the field "exclude_from_intrastat_if_present" on account.tax.template (it only has it on account.tax).

So, in order to avoid loo much hassle, I'll probably delete the 12.0 branch and report from v11 with good code.

alexis-via:
OK, my mistake, the field "exclude_from_intrastat_if_present" on "account.tax.template" is not on origin/11.0... I need to understand why, because I have it in l10n_fr_intrastat_product in OCA/l10n-france for v11 cf https://github.com/OCA/l10n-france/blob/11.0/l10n_fr_intrastat_product/data/account_tax_template.xml
Very very strange...

PR #70, opened 2019-05-28
Migrate all modules from v11 to v12

alexis-via:


moylop260:
Could you check the following post, please?
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0

rvalyi:
Hello @moylop260 
I was wondering if this was still the preferred way in the OCA (we always did it this way) as recently contributors such as @sebastienbeau and @sbidoul concluded the old branch way was making it easier to merge PRs in either branches when the code doesn't differ too much as it messes less with the history. I didn't follow if something was decided about this in the last meeting or something, all I know is that it's the way @sebastienbeau and @sbidoul started the 12.0 branches of the Shopinvader repos for instance (I'm not questionin the advantages of the reset way that helped clean legacy modules out from the early OCA days). So I'm just asking: are you sure the link you mentionned is still the required way?

moylop260:
Hello @rvalyi 
Let me ask to OCA members board
@max3903 @pedrobaeza 
Are there changes in the way to migrate modules different to https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0?

rvalyi:
@moylop260 in doubt: what is the thing that annoys you really here? Is it the way the 12.0 branch was started or is it the per-module PR rule, or both?

moylop260:
I saw 68 commits for 11.0 branch:
 - https://github.com/OCA/intrastat-extrastat/commits/11.0

It PR talk about migrate all modules but it has just 4 commits.
I expected saw >68 commits to migrate to 12.0

Maybe I'm missing something special for this report.

rvalyi:
Hello @moylop260 
I'm not trying the defend the way the migration is done here, but may be there is a misunderstanding: this is not the usual case were the guy screwed the project history. In fact the project history is already present in the 12.0 branch because the branch was created the old way like

```
git checkout 11.0
git checkout -b 12.0
```
then, like during the 6.1->7.0 and 7.0->8.0 migrations the modules were set as ```'instalable': False```
https://github.com/OCA/intrastat-extrastat/commit/2efd611708ff054088d35f59e0f79d7a78f7e2db

And so these 4 commits here are really just the migration commits, albeit not PR module as expected (yet these are simple changes).

In case one still think this is a problem I can assist to rebuild the 12.0 branch and do the cherry-pick/migration PRs. Let me know.

pedrobaeza:
@moylop260 do you have it clear or need extra explanations?

I have to say that I don't like too much to have several modules migrated at the same time in the same PR, but I'm not going to block it as I don't contribute directly to this repo.

PR #96, opened 2020-04-14
[11.0][FIX] 11.0 add intrastat product

RodrigoBM:
To whom it may concern,

From  _FactorLibre_ we have seen the need to modify the way we extract the product lines from the invoices before they are provided to Intrastat. Instead of extracting the product lines by searching the invoices one at a time, we have made it so the lines are obtained by searching all product lines within all invoices, in accordance to the following criteria:

If the declaration being generated is a “Declaration for Arrival”, only the lines that comply with the following criteria will be taken into account, since they imply a product entry:

- Lines from customer invoices in which the quantity is negative.
- Lines from customer rectification invoices in which the quantity is positive.
- Lines from suppliers invoices in which the quantity is positive.
- Lines from suppliers rectification invoices in which the quantity is negative.

If the declaration being generated is “Declaration for Dispatches”, only the lines that comply with the following criteria will be taken into account, since they imply a product departure:

- Lines from customer invoices in which the quantity is positive.
- Lines from customer rectification invoices in which the quantity is negative.
- Lines from suppliers invoices in which the quantity is negative.
- Lines from suppliers rectification invoices in which the quantity is positive.

The main reason for this modification is being able to have into account the invoices generated by a refund or the exchange of a product for another. This generates a rectification invoice that displays a line in which the quantity is negative and a line in which the quantity is positive. Our goal is to allow the system to distinguish which line corresponds to an entry and which to a departure when they’re added to the Intrastat report.

Thank you for your time.

Best regards.

### DIFF
--- a/intrastat_product/README.rst
+++ b/intrastat_product/README.rst
@@ -87,6 +87,7 @@ Contributors
 
 * Alexis de Lattre, Akretion <alexis.delattre@akretion.com>
 * Luc De Meyer, Noviat <info@noviat.com>
+* Rodrigo Bonilla, FactorLibre <rodrigo.bonilla@factorlibre.com>
 
 Maintainer
 ----------
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -253,7 +253,7 @@ class IntrastatProductDeclaration(models.Model):
             note += "\n" + _(
                 "Please adjust this line manually.")
             self._note += note
-            return weight, suppl_unit_qty
+            return abs(weight), abs(suppl_unit_qty)
 
         if intrastat_unit_id:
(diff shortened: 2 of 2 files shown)


luc-demeyer:
@RodrigoBM 
I am not sure if we should go this way.
Imho a refund should be used when returning goods in stead invoice lines with negative quantities.
In the belgian legislation, arrivals because of returning goods need to declared differently as regular arrivals, cf. https://github.com/Noviat/noviat-apps/blob/ca1cc96ae0d4cc64475c9b2c743173c173db444b/l10n_be_intrastat_product/models/l10n_be_intrastat_product_declaration.py#L48.
Changing the logic as you suggest would break the belgian localisation module.
I have no view on impact on e.g. french localisation.

@alexis-via 
what's your opinion on htis

alexis-via:
@luc-demeyer I agree with you

Issue #97, opened 2020-05-26
Problem  [12.0] [intrastat_base]


matteoopenf:
Hi there,
when trying to enter the general settings of Odoo, I get this error:

```
Odoo Server Error
Traceback (most recent call last):
  File "/opt/odoo/odoo/odoo/api.py", line 1049, in get
    value = self._data[key][field][record._ids[0]]
KeyError: <odoo.models.NewId object at 0x7f5cc17c9d98>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/odoo/fields.py", line 1004, in __get__
    value = record.env.cache.get(record, self)
  File "/opt/odoo/odoo/odoo/api.py", line 1051, in get
    raise CacheMiss(record, field)
odoo.exceptions.CacheMiss: ('res.config.settings(<odoo.models.NewId object at 0x7f5cc17c9d98>,).intrastat_remind_user_ids', None)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/odoo/api.py", line 1049, in get
    value = self._data[key][field][record._ids[0]]
KeyError: 1

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/odoo/fields.py", line 1004, in __get__
    value = record.env.cache.get(record, self)
  File "/opt/odoo/odoo/odoo/api.py", line 1051, in get
    raise CacheMiss(record, field)
odoo.exceptions.CacheMiss: ('res.company(1,).intrastat_remind_user_ids', None)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/odoo/http.py", line 656, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/odoo/odoo/odoo/http.py", line 314, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/opt/odoo/odoo/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/opt/odoo/odoo/odoo/http.py", line 698, in dispatch
    result = self._call_function(**self.params)
  File "/opt/odoo/odoo/odoo/http.py", line 346, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/odoo/odoo/odoo/service/model.py", line 97, in wrapper
[...]

PR #98, opened 2020-05-29
change m2m table name to avoid naming conflicts with other modules

luc-demeyer:
cf. https://github.com/OCA/intrastat-extrastat/issues/97

### DIFF
--- a/intrastat_base/models/res_company.py
+++ b/intrastat_base/models/res_company.py
@@ -10,7 +10,9 @@ class ResCompany(models.Model):
     _inherit = "res.company"
 
     intrastat_remind_user_ids = fields.Many2many(
-        'res.users', column1='company_id', column2='user_id',
+        comodel_name='res.users',
+        relation='company_intrastat_reminder_user_rel',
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
You need to add a migration script that rename previous table, or does that table already exists before this module?

matteoopenf:
> You need to add a migration script that rename previous table, or does that table already exists before this module?

I think is not possible in this specific  case

matteoopenf (approved):

luc-demeyer:
@pedrobaeza 
I agree that in principle a migration script would be needed but in reality I have no single customer using this option (hence no impact) and I doubt that there are installation having more than 1 or 2 of such users (hence spending time on such a migration script is not high on my priority list but I agree it's also not a lot of work).

pedrobaeza:
OK, me neither, but I prefer to be totally correct, but if you think we should go ahead, then I refrain my review.

alessandrocamilli (approved):

matteoopenf:
Hi @alexis-via merge?

PR #100, opened 2020-06-08
[13.0] intrastat_product: Add xml_attachment_id field and various improvements

alexis-via:
intrastat_product: add mail.activity.mixin and widget in form view
product_harmonized_system: remove archive button (use Action menu)
All modules:
- use super() instead of super(class, self)
- Update domain_force in multi-company ir.rule to v13 standard

### DIFF
--- a/intrastat_base/models/intrastat_common.py
+++ b/intrastat_base/models/intrastat_common.py
@@ -148,7 +148,7 @@ class IntrastatCommon(models.AbstractModel):
                     _("Cannot delete the declaration %s " "because it is in Done state")
                     % self.year_month
                 )
-        return super(IntrastatCommon, self).unlink()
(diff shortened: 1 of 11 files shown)


luc-demeyer (approved):
Code review  & testing ok.
Nice work, imho we should merge this one asap.
Adding a migration script for the company record rule would be a usefull but this is not blocking for me.

pedrobaeza (changes requested):
The change in record rules requires migration script. See https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0#tasks-to-do-in-the-migration for an example

alexis-via:
Here they are!

pedrobaeza (approved):

PR #101, opened 2020-08-01
[13.0]allow to encode intrastat transaction data on invoice line without product

luc-demeyer:
We get a stack trace on the intrastat report when entering an invoice line with an intrastat code but without product code.
This PR fixes this issue and allows at the same the users to add/modify the weight and the country of origin on the invoice line.
This use case is common when e.g. receiving a supplier invoice with intrastat codes and weight.
Creating a product record for a on-time purchase doesn't make a lot of sense.
Some companies also trade goods where the weight in different per transaction, this is now also fixed.

### DIFF
--- a/intrastat_product/__manifest__.py
+++ b/intrastat_product/__manifest__.py
@@ -21,6 +21,8 @@
     ],
     "excludes": ["account_intrastat"],
     "data": [
+        "security/intrastat_security.xml",
+        "security/ir.model.access.csv",
         "views/hs_code.xml",
         "views/intrastat_region.xml",
         "views/intrastat_unit.xml",
@@ -31,8 +33,6 @@
         "views/account_move.xml",
         "views/sale_order.xml",
         "views/stock_warehouse.xml",
-        "security/intrastat_security.xml",
-        "security/ir.model.access.csv",
         "data/intrastat_transport_mode.xml",
         "data/intrastat_unit.xml",
     ],
--- a/intrastat_product/__manifest__.py
+++ b/intrastat_product/__manifest__.py
@@ -7,11 +7,11 @@
 
 {
     "name": "Intrastat Product",
-    "version": "13.0.1.0.3",
+    "version": "13.0.1.1.0",
     "category": "Intrastat",
     "license": "AGPL-3",
     "summary": "Base module for Intrastat Product",
-    "author": "brain-tec AG, Akretion, Noviat, " "Odoo Community Association (OCA)",
+    "author": "brain-tec AG, Akretion, Noviat, Odoo Community Association (OCA)",
     "depends": [
         "intrastat_base",
         "product_harmonized_system",
--- a/intrastat_product/models/account_move.py
+++ b/intrastat_product/models/account_move.py
@@ -99,6 +99,21 @@ class AccountMove(models.Model):
         return vals
 
 
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    hs_code_id = fields.Many2one(
+        comodel_name="hs.code", compute="_compute_hs_code_id", string="Intrastat Code",
+    )
+
+    def _compute_hs_code_id(self):
+        for rec in self:
+            intrastat_line = self.move_id.intrastat_line_ids.filtered(
+                lambda r: r.invoice_line_id == rec
+            )
+            rec.hs_code_id = intrastat_line.hs_code_id or rec.get_hs_code_recursively()
+
+
 class AccountMoveIntrastatLine(models.Model):
     _name = "account.move.intrastat.line"
     _description = "Intrastat declaration details"
--- a/intrastat_product/models/account_move.py
+++ b/intrastat_product/models/account_move.py
@@ -84,15 +84,18 @@ class AccountMove(models.Model):
 
     def _get_intrastat_line_vals(self, line):
         vals = {}
-        if self.env["intrastat.product.declaration"]._is_product(line):
+        decl_model = self.env["intrastat.product.declaration"]
+        if decl_model._is_product(line):
             hs_code = line.product_id.get_hs_code_recursively()
             if not hs_code:
                 return vals
+            weight, qty = decl_model._get_weight_and_supplunits(line, hs_code)
             vals.update(
                 {
                     "invoice_line_id": line.id,
                     "hs_code_id": hs_code.id,
-                    "transaction_weight": int(line.product_id.weight * line.quantity),
+                    "transaction_weight": int(weight),
+                    "transaction_suppl_unit_qty": qty,
                     "product_origin_country_id": line.product_id.origin_country_id.id,
                 }
             )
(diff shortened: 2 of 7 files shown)


intrastat_product/__manifest__.py
```diff
@@ -7,11 +7,11 @@
 
 {
     "name": "Intrastat Product",
-    "version": "13.0.1.0.3",
+    "version": "13.0.1.1.0",
```
pedrobaeza, line 10: Please next time don't change version number, as it's done by the bot when merging and you avoid possible conflicts.

intrastat_product/models/account_move.py
```diff
@@ -67,9 +67,16 @@ class AccountMoveLine(models.Model):
     hs_code_id = fields.Many2one(
         comodel_name="hs.code", string="Intrastat Code", ondelete="restrict"
     )
+    weight = fields.Integer(string="Weight", help="Net weight in Kg")
```
pedrobaeza: Why do you need these extra fields? I think it's too much, and more not being even computed.

luc-demeyer: @pedrobaeza 
They should indeed not being computed since this is to encode an invoice without product records (hence only intrastat code, weight & country of origin) or in case of product records with unknown weight (product with variable weight, hence weighting operation needed).

pedrobaeza: But I don't get why this fixes the stack trace.

luc-demeyer: @pedrobaeza 

The stack trace is fixed by the "if not product::" code block (cf. PR line 275 of the declaration)

pedrobaeza: Then please limit this PR to that code, an propose improvements into another for debating such improvement.

luc-demeyer: Not really in this case since the fix would be completely different (disallow entering intrastat code without product record). This feature should have been there for a long time since (at least on our market) common accounting packages support this. With the optional="hide" in Odoo 13.0 we can do this in a rather clean way.

pedrobaeza: But it doesn't make sense to put something without product. For me it should be required if you want Intrastat.

luc-demeyer: Oh yes it does.

Assume you are importing on a regular basis goods from EU countries which makes your company subject to the intrastat arrival declaration.
For the goods that you are importing on a regular basis you will work with product records, purchase orders, ...
But for some other goods you may work without purchase orders or a generic product record.
If you receive your supplier invoice it is a common practice that the intrastat code is on the invoice lines (if not than you need to take the shipping docs).
You don't want to do the overhead of creating a product record with weight for a one time purchase hence you simply encode the intrastat code and weight on the invoice line.
Especially for companies migrating from an accounting package having this feature, this is a must have (this particular PR is driven by a customer who migrated from winbooks, which is the leading accounting package on the belgian market).

The other scenario is for products with variable weight. The product code is always the same but the weight is different per transaction. Hence you need to take the shipping docs when encoding the invoice and enter the weight in order to generate your intrastat declaration.

pedrobaeza: And why this hasn't been a problem till v13? As said, this is an improvement, and not a bug fix, and it should be debated in a separate PR. We have to avoid the stack trace, and the bugfix is to make the product required. 

All the rest for me should be debated in a separate PR, but you already know that I don't want to bring such low level data to each account move where inconsistency is going to happen for sure:

- Incorrect data in the product, that is propagated to lines
- On previous versions, you fix it on the product, and everything is OK. Not with this change.

luc-demeyer:
I changed this PR description from a bug fix to an enhancement (which it is but as a side effect this enhancement fixes also a bug).

>>> And why this hasn't been a problem till v13
It has always been a problem, only now it has become a blocking issue for a concrete customer project.

>>> Incorrect data in the product, that is propagated to lines
Without this enhancement you may have to create a thousand product records for the same product since at any sale/purchase transaction the weigth is different. With this enhancement we allow to encode intrastat data without product record. 
I agree that adding more fields to the account.move.line increases the risk on conflicts but another database model (e.g. create a separate record with m2o link to the invoice line) doesn't look very attractive to me either.

Now that the title of this PR is corrected we can keep this record to have the debate on this feature and based upon this feedback we may change the way we implement it from an UI experience, database model to store the extra fields, implement it in a separate module, ...

pedrobaeza:
OK, my vote anyway is to not add this in the base: the overhead and inconsistencies that will provoke is not worth for that use case. Forcing to create a unique product is reasonable IMO if you have to do Intrastat. I would accept in order of preference:

1. Having weights editable at Intrastat report level, so those not putting weights on product (generic products), can do it here. I know that the answer to this is that they know the weights on the moment the invoice is issued, not on the summary, and I can agree.
2. To not propagate weights from product in onchange, so the priority is always in the product, and only manual typed weights are taken into account. This way we avoid inconsistencies and to fill DB with the same data again and again without need. This is not very UX either. We can have a computed/inverse field for showing the product value, and only when changed, stored the weight in another field.

In fact, the same should apply to HS code, that is being now propagated without need (but I'm late to party in that decision), but an homogeneous treatment in this would be also interesting, or at least remove the onchange as option 2.

luc-demeyer:
@pedrobaeza 

Thanks for the feedback.
I'll try to find a bit of time this week to rework this PR by avoiding the hscode & weigth propagation but still allow the user to modify them.

Since we already have invoice lines and move lines we could also add an 'intrastat lines' notebook page showing the computed/related hscode & weight fields and allowing the user to enter transaction specific fields (called e.g. "transaction_weight" & "transaction_hscode"), hence no propagation, no onchange.

The intrastat report will first look for the "transaction_*" fields and fall back on the product record defaults in case they are empty.

And we keep of course the current capability to adjust the lines in the declaration.

Remains the decision to put this into a separate module or not.
Adding it to the "intrastat_product" module is less work.
From a UI standpoint we can easily hide the "intrastat_lines" notebook page with a useability group.

pedrobaeza:
Thanks for taking into account my comments.

luc-demeyer:
@pedrobaeza 
New version is now ready. 
All account.move.line fields have gone (including the hs_code_id which has been there always).
A usability group is required to encode intrastat transaction details on the invoices.

I have not added a migration script on the hs_code_id field since it would be hard to do (the new concept with the transaction details is completely different) and I don't think it's needed since the declaration history is already kept in the declarations.

If someone believes it is important I could rename the hs_code_id field in the account_move_line table to e.g. hs_code_id_legacy via a pre-migration script.

@alexis-via 
Can you also have a look ?

pedrobaeza:
There's a weird warning in runbot:
```
2020-08-18 21:30:16,258 158 WARNING openerp_test odoo.addons.base.models.ir_ui_view: The group intrastat_product.group_invoice_intrastat_transaction_details defined in view intrastat.invoice.form does not exist!
```

I have checked code quickly and seems OK. Thanks for the efforts.

luc-demeyer:
@pedrobaeza 
runbot warning now fixed.

pedrobaeza:
@luc-demeyer please don't self-merge when the PR is so dirty, with merge commits and so on, that messes up commit history when migrating to upper versions. It's not the first time this happens. I'm trying to fix it force pushing...

PR #105, opened 2020-10-14
[14.0][MIG] intrastat_base

alexis-via:


rvalyi (approved):

luc-demeyer (changes requested):
I propose we align the intrastat modules fully to the OCA guidelines as part of the 14.0 migration.
Imho not a lot of work for this one.
- rename views to comply with "views/<main_model>_views.xml" file naming convention
- move res_config_settings to wizards folder

alexis-via:
I'm not really in favor of strictly adopting the OCA guidelines, in particular when they don't make a lot of sense. For example, what's the point of adding a "_views.xml" suffix when you are in the "views" subfolder ? "views/<main_model.xml" makes more sense and avoids 6 caracters :)

luc-demeyer:
@alexis-via 
It's not a breaking point to me for merging this PR but I also don't think we should be solo players in this repo when we disagree with some OCA guidelines. Such a discussion should be initiated on the maintainer tools repo.

Concerning this specific guideline:
When looking into e.g. https://github.com/odoo/odoo/tree/14.0/addons/account/views than you see there is a point, odoo terminates nearly all xml files in this folder with _views but account_menuitem.xml is different since no views in this file.

bealdav:
see https://github.com/OCA/maintainer-tools/pull/466

PR #111, opened 2020-12-11
[13.0][FIX] Migration fixes + full migration scripts

MiquelRForgeFlow:
Please, review + merge ASAP :)

### DIFF
--- a/intrastat_product/migrations/13.0.1.0.3/post-migration.py
+++ b/intrastat_product/migrations/13.0.1.0.3/post-migration.py
@@ -2,9 +2,69 @@
 
 from openupgradelib import openupgrade  # pylint: disable=W7936
 
+_months = [
+    (1, "01"),
+    (2, "02"),
+    (3, "03"),
+    (4, "04"),
+    (5, "05"),
+    (6, "06"),
+    (7, "07"),
+    (8, "08"),
+    (9, "09"),
+    (10, "10"),
+    (11, "11"),
+    (12, "12"),
+]
+
+
+def map_intrastat_product_declaration_month(env):
+    openupgrade.map_values(
+        env.cr,
+        openupgrade.get_legacy_name("month"),
+        "month",
+        _months,
+        table="intrastat_product_declaration",
+    )
+
+
+def update_invoice_relation_fields(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_move am
+        SET (intrastat_transaction_id, intrastat_transport_id,
+            src_dest_country_id, intrastat_country, src_dest_region_id
+            ) = (ai.intrastat_transaction_id,
+            ai.intrastat_transport_id, ai.src_dest_country_id,
+            ai.intrastat_country, ai.src_dest_region_id)
+        FROM account_invoice ai
+        WHERE am.old_invoice_id = ai.id""",
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_move_line aml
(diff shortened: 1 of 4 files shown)


JordiMForgeFlow (approved):

LoisRForgeFlow (approved):

intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -477,11 +477,12 @@ def _prepare_invoice_domain(self):
         start_date = date(int(self.year), int(self.month), 1)
         end_date = start_date + relativedelta(day=1, months=+1, days=-1)
         domain = [
-            ("date", ">=", start_date),
-            ("date", "<=", end_date),
+            ("invoice_date", ">=", start_date),
```
pedrobaeza, line 480: I think the accounting date is used here on purpose

intrastat_product/migrations/13.0.1.0.3/post-migration.py
```diff
@@ -2,9 +2,69 @@
...
+        table="intrastat_product_declaration",
+    )
+
+
+def update_invoice_relation_fields(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_move am
+        SET (intrastat_transaction_id, intrastat_transport_id,
```
pedrobaeza, line 36: Is this syntax allowed instead of `field1 = value1, field2 = value2`?

intrastat_product/models/intrastat_product_declaration.py
MiquelRForgeFlow: In version 12, it was date_invoice

pedrobaeza: Ok then

intrastat_product/migrations/13.0.1.0.3/post-migration.py
MiquelRForgeFlow: see in https://www.postgresql.org/docs/9.5/sql-update.html, there is an example saying "Use the alternative column-list syntax to do the same update:"

pedrobaeza: OK, TIL

luc-demeyer:
I think this one got merged a little bit too fast.
For the amount calculation in multi-currency environments the code is now based upon the "invoice_date" field whereas Odoo will use the "date" field for the the currency conversion.
In Belgium the tax authorities perform a cross control between the Intrastat and Itracommunity declaration. If they do not match, you get a control by the authorities which is something everyone wants to avoid.
Hence we should change the invoice.invoice_date to invoice.date.

pedrobaeza:
I have asked about that, but as @MiquelRForgeFlow commented, in v12 it was using invoice date.

MiquelRForgeFlow:
Hi @luc-demeyer The correspondence is as following:

invoce.date -> move.date
invoice.date_invoice -> move.invoice_date

So, it using here invoice_date is wrong, then in v12 also is wrong using date_invoice.

luc-demeyer:
It looks like this has always been wrong.
I also know that my customers adapt the intrastat declaration when they see small differences between the tax and intrastat declaration, hence they probably never reported this issue. 
I'll make a PR for Odoo 13.0 to fix it so that we move this fix forward into 14.0.

pedrobaeza:
It can be included by @MiquelRForgeFlow in #112. We should fix it then on 12.0 as well.

luc-demeyer:
cf. https://github.com/OCA/intrastat-extrastat/pull/113

PR #113, opened 2020-12-11
[13.0]invoice_date -> date

luc-demeyer:
Intrastat declaration should match intracomm tax declaration, hence we need to use the accounting date for the intrastat declaration.

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -350,7 +350,7 @@ class IntrastatProductDeclaration(models.Model):
             inv_line.price_subtotal,
             self.company_id.currency_id,
             self.company_id,
-            invoice.invoice_date,
+            invoice.date,
         )
         return amount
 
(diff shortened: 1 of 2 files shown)


intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -477,8 +477,8 @@ def _prepare_invoice_domain(self):
         start_date = date(int(self.year), int(self.month), 1)
         end_date = start_date + relativedelta(day=1, months=+1, days=-1)
         domain = [
-            ("invoice_date", ">=", start_date),
-            ("invoice_date", "<=", end_date),
+            ("date", ">=", start_date),
```
pedrobaeza, line 482: Are you sure these domains should be by accounting date? One thing is the amount, and the other is in which period to include the invoices.

luc-demeyer:
If the intrastat deviates from the tax declaration than you may be in trouble (although I am not how the data mining works in the different EU member states).

https://ec.europa.eu/eurostat/documents/3859598/8512202/KS-07-17-102-EN-N.pdf/736c4a50-c240-4144-b087-4fa6aece8ee0
In this publication it is said that a member state may define phyiscal movement date or tax declaration date.
I think that in real life most software package are based upon accounting dates and not on phyiscal move dates and I am not in favor of adding an extra complexity to this module to base the logic on stock.move date with extra config parameter.

pedrobaeza:
OK, then let's continue with that approach, but please add this as comment in the code where dates are used for later references and avoid these doubts in the future, don't you think?

pedrobaeza (approved):

Issue #110, opened 2020-12-03
14.0 mig : invoice domain should be based on fiscal position in stead of country


luc-demeyer:
Up till now the default domain to select intrastat transactions is based upon invoices to/from EU countries.
I propose that we change in 14.0 the logic to be based upon the "Intra EU" fiscal position.

On the Belgian market transactions subject to the 'intrastat' declaration are also subject to the 'intracommunity' tax declaration.
In the 13.0 version of the Belgium localisation module I already replaced the domain query, cf. 
https://github.com/Noviat/noviat-apps/blob/012c9a254b101c405126b1c0ce5276a0a9c16d9e/l10n_be_intrastat_product/models/l10n_be_intrastat_product_declaration.py#L219

But since these rules are decided on an european level, this should be applicable for all EU countries.
As a consequence I propose that for the 14.0 migration we add an 'intrastat' boolean to the fiscal position and base the domain on invoices with fiscal positions marked as 'intrastat'. In the localisation modules we can add the logic to set this flag automatically on the standard Intra EU fiscal position.

To give a concrete example why the country tabel is wrong:
If you ship goods to a customer without VAT number in an EU country that this shipment must not be included in the intrastat declaration.

alexis-via:
I agree with this idea.
About your last point, the scenario was handled via the boolean exclude_from_intrastat_if_present on account.tax and it was working fine.
The advantage of this solution is that it's easier to configure, because you just a few fiscal positions instead of dozens of taxes.

alexis-via:
Now that this has been implemented in v14, let's close it.

PR #115, opened 2021-01-05
[14.0] intrastat_base: use fiscal position

alexis-via:
Please wait for the PR of the migration of intrastat_product to v14 before merging, because I may need to introduce other changes.


luc-demeyer (changes requested):
please adapt USAGE.rst to reflect changed fields (fiscal postion boolean i.s.o. country boolean and flag on taxes)

luc-demeyer (approved):

pedrobaeza:
I don't get why it's better to remove the flag on countries and put it on fiscal positions. This makes the need to add a localization module to each country and see if there's a FP that matches the requirement. Can you explain the reasons?

luc-demeyer:
@pedrobaeza : it's explained in issue https://github.com/OCA/intrastat-extrastat/issues/110

pedrobaeza:
I see, but the problem I depicted is still there, so I propose a mixed formula: keep country flag, but give more priority to fiscal position one.

luc-demeyer:
Keeping the country flag doesn't make a lot of sense since the query on the fiscal position is one that gives the correct result.
A query on country may result in deviating intrastat and tax reports and usually the tax reports are the correct ones.
A country flag fallback logic will make the whole thing imho only more confusing and increase risk on misconfiguration.
The localisation modules require a small change if they want to set the intrastat flag on the fiscal position but without this change if will still work if the user reads the intrastat "README.rst" and configures this flag manually.

pedrobaeza:
It's not a small change, as this impact on both existing installations + new, as the fiscal position is generated for each company. I prefer this change to not be applied, or not without explicit opt-in, and let the previous one as fallback.

luc-demeyer:
@pedrobaeza 
technically it is easy to add the boolean for each company in a migration script:

for company in companies:
    fpos_ref = "{}.{}_{}".format(module.name, .company.id, fpt_name)
   ...

companies: all spanish companies in a database.
module = 'l10n_es'
fpt_name = xml_id name of the intra community fiscal position

If you leave the country flag than you hit the problem of wrong declaration entries for e.g. B2C customers in those countries.
If we remove it completely and your customer has created his own fiscal position (and not the one from the localisation module) than he will see an empty declaration and realise that something is wrong (hence call his integrator or read the module info and set the flag).

Haviing a simple boolean on the intracomm fiscal position is the easiest way to configure this and at the same time it handles the brexit case. 
It is a change but a lot smaller than the type of changes made by Odoo when they create a new version.

You can of course still add a country-based fallback in the spanish localisation module and maintain the table of EU-intrastat countries at that place but I don't recommend since it does not cover the brexit case and it adds imho only complexity and confusion.

@alexis-via 
What do you think ?

pedrobaeza:
OK, it's not ideal due to what I have commented, but I understand the use case, and I will try to add at Spanish Intrastat glue module the automatic checking of the mark for the proper Spanish fiscal position.

/ocabot merge major

alexis-via:
Seems the merge bot is dead. How should we proceed @pedrobaeza ?

PR #116, opened 2021-01-05
[14.0][MIG] intrastat_product

alexis-via:


joao-p-marques (commented):
Code review :+1: 

The tests seem to be failing due to `stdnum`. The support for the new Brexit VAT numbers has been added in https://github.com/arthurdejong/python-stdnum/commit/53f13b4af087250e43f8841c01906e46d5687b44 and has not landed in a release yet, so we need to install from that specific commit (I have done this locally and the tests pass) or wait for the release :man_shrugging: 

The rest seems good to me. Can you please just squash the commits and put the appropriate suffix (`[MIG]`, `[IMP]`, etc.)?

pedrobaeza:
There's no black, prettier, isort commit as well.

joao-p-marques:
Yes, I guess it all went in the `[MIG] intrastat_product to v14` commit. @alexis-via can you separate it according to the [guidelines](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0)?

alexis-via:
This PR is a migration from v13, so the code of the module intrastat_product was already formatted with black/isort/...

pedrobaeza:
There are changes that always happen from one version to another (example: setup folder), and we also bump the version of black and the rest of the libraries involved, so please follow the exact steps on the migration guide: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0

intrastat_product/models/hs_code.py
```diff
@@ -0,0 +1,37 @@
...
+class HSCode(models.Model):
+    _inherit = "hs.code"
+
+    intrastat_unit_id = fields.Many2one(
+        comodel_name="intrastat.unit", string="Intrastat Supplementary Unit"
+    )
+
+    @api.constrains("local_code")
+    def _hs_code(self):
+        eu_countries = self.env.ref("base.europe").country_ids
```
luc-demeyer, line 19: I don't think it's a good idea to base a constraint on base.europe since the data in that country group is not guaranteed to be correct (noupdate=1).
I suppose the reason behind this is that some non-eu countries extend the hs_code with 'non-digits'.

I propose we simply drop the isdigits check here and move the check to the generation of the declaration.

pedrobaeza: Is Intrastat used outside Europe? If not, why make the check only in this case? Let's put the constraint applied universally.

pedrobaeza:
Superseded by #120 as Alexis is not very friend of git operations. We can continue here the open discussion about the constraint if you prefer, and we will reflect it in the other PR.

intrastat_product/models/hs_code.py
luc-demeyer: @pedrobaeza 
Intrastat is within europe (+nothern ireland) but hs_code is international hence we shouldn't add constraints breaking use of hs_code outside europe. That why I propose to remove the constraint here and add it as a check when generating the declaration.

pedrobaeza: OK, I see. But reading https://en.wikipedia.org/wiki/Harmonized_System, it seems this classification is normalized across the world, and there aren't specific Europe constraints, isn't it?

pedrobaeza: Another thing that I'm thinking of is that `product_harmonized_system` should be in OCA/product-attribute or OCA/community-data-files then.

luc-demeyer: @pedrobaeza 
1) european constraints: the european countries extend the number with digits. I am not sure if the rest of the world extend the codes with digits or characters. @rvalyi : do you have a better view on this ?

2) product-attribute
I am not sure. This repo is called intrastat-extrastat since it goes beyond europe.
For a recent project with import for Asia we have done some work to extend the HS Codes with a country table and add the country specific duties. This work in intended to end up in this repo as well.

pedrobaeza: OK, then just let's remove them for now and we decide later. At the end, for most of us, the codes are imported via a wizard, so it's not a great problem. cc @joao-p-marques

PR #118, opened 2021-01-22
[10.0] intrastat_base: Brexit (simple, without support for Northern Ireland)

alexis-via:
intrastat = False on UK, with update via mig script
I don't think we'll backport all the work for Northern Ireland we made in v14...
@luc-demeyer If you are OK with this PR, I'll do the same for v12 and v13.

### DIFF
--- a/intrastat_base/__manifest__.py
+++ b/intrastat_base/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Intrastat Reporting Base',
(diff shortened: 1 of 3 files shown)


alexis-via:
@luc-demeyer  are you OK with this ? If yes, I'll merge it and i'll do the same with other odoo versions (v12, v13, ...)

luc-demeyer (approved):

PR #119, opened 2021-01-27
[13.0][IMP] Move Country of Origin to product variants

carlosserratoro:


product_harmonized_system/__manifest__.py
```diff
@@ -9,7 +9,7 @@
 
 {
     "name": "Product Harmonized System Codes",
-    "version": "13.0.1.0.1",
+    "version": "13.0.1.0.2",
```
sebalix, line 12: Same remark than https://github.com/OCA/community-data-files/pull/104/files#r565440351

product_harmonized_system/views/product_product.xml
```diff
@@ -0,0 +1,36 @@
...
+<odoo>
+    <record id="product_variant_easy_edit_view" model="ir.ui.view">
+        <field name="name">product.product.view.form.easy</field>
+        <field name="model">product.product</field>
+        <field name="inherit_id" ref="product.product_variant_easy_edit_view" />
+        <field name="arch" type="xml">
+            <sheet position="inside">
+                <group
+                    name="info"
+                    string="General Information"
```
sebalix, line 11: High chance to have another group named `info` (quite used) by another module, is there no other existing page/group to use to add this field? Or maybe rename this group?

luc-demeyer: I propose to change this to name="intrastat-extrastat"

product_harmonized_system/views/product_product.xml
```diff
@@ -0,0 +1,36 @@
...
+                            attrs="{'invisible': [('type', '=', 'service')]}"
+                        />
+                    </group>
+                </group>
+            </sheet>
+        </field>
+    </record>
+    <record id="product_normal_form_view" model="ir.ui.view">
+        <field name="model">product.product</field>
+        <field name="inherit_id" ref="product.product_normal_form_view" />
```
luc-demeyer, line 26: Afaik product_normal_form_view inherits from the product.template form view.
In this view we already add the Country of Origin. Maybe I am missing something here.

product_harmonized_system/models/product_template.py
```diff
@@ -20,16 +20,17 @@ class ProductTemplate(models.Model):
         "http://www.wcoomd.org/. You can leave this field empty "
         "and configure the H.S. code on the product category.",
     )
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
```
luc-demeyer, line 27: I am not against moving the origin_country_id to the variant but I think we should in that case consider to really move this field, not an extra field on top of the product.template since this could be hard to manage.
Why not turn the origin_country_id  on product.template into a computed field like odoo does with the default_code field, cf. https://github.com/odoo/odoo/blob/1b57cf252dc868c3b4ff59894fad39f2e955dab5/addons/product/models/product_template.py#L146

carlosserratoro:
Closed to not break existing functionality.

PR #120, opened 2021-02-08
[14.0][MIG+IMP] intrastat_product: Migration to 14.0

joao-p-marques:
Supersedes https://github.com/OCA/intrastat-extrastat/pull/116

Squash commits and add appropriate prefix. Also add `[IMP]` commit for pre-commit execution.

@Tecnativa
TT28130

ping @pedrobaeza


joao-p-marques:
Tests should not give the same error as before in the superseded PR as the latest release of `python-stdnum` already includes the Brexit support.

pedrobaeza:
You can squash these commits together:
Selección_027
and these ones:
Selección_028

joao-p-marques:
Done @pedrobaeza

pedrobaeza:
Why this commit dfc322717c11c6c7773c00fa1361d0f26438a958 contains translations?

I think you have squash them, when in fact it should be squashed with ee61d1122b01dfb2faffad2fc6e7689f68cc5b1a. You can squash all together, coming from same author, and modifying the commit message for containing both messages.

joao-p-marques:
Yes, indeed I squashed these:
image

Am I supposed to squash the bot commits all in 1 also authored by the bot? I thought they should be squashed with the previous commit that triggered that change by the bot.

I will squash it with the previous one, as you said.

pedrobaeza:
Yes, but you should see a bit the context. The order that you see is not always the sequential order. It can be due to several things:

- A merge contains several commits, and one of them is the "generator" of the rest of the changes. This shouldn't be a problem if all comes from the same author, as in this context, you can do what you are going to do now: squash all together, mixing the commits messages.
- The PR was done some time ago from the merge of it, and then in between other PRs/Weblate commits/etc may arrive in the main branch, provoking that disorder.

joao-p-marques:
Ok @pedrobaeza thanks for the explanation. In this case, I only squashed the commits referring to the v14.0 migration, the administrative commits, and the ones you pointed out. Do I need to do anything else or revert anything?

pedrobaeza:
Commit history seems good. Please check Travis

joao-p-marques:
`pre-commit` is generating an incorrect `requirements.txt` file after I have added the `>=1.16` constraint: https://github.com/OCA/intrastat-extrastat/pull/120/checks?check_run_id=1855987879#step:4:94. Am I doing it wrong? Where should that constraint go?

pedrobaeza:
I think you have to put that in external_dependencies in Odoo, with pip syntax.

joao-p-marques:
Yes, I did that: https://github.com/OCA/intrastat-extrastat/pull/120/commits/ec3752d2ff87156fb1457a3b963fa63b0f816e89#diff-66b4fe70725cdf32dd709c35ce332f272cbf58b4e20ae98ad6ed59a3ed1d6763R24

Before that, pre-commit generated the correct `requirements.txt` with `python-stdnum`, but now generates `stdnum>=1.16`.

pedrobaeza:
@sbidoul @Yajo can you help here?

sbidoul:
Just put `python-stdum>=1.16" [here](https://github.com/OCA/intrastat-extrastat/pull/120/files#diff-66b4fe70725cdf32dd709c35ce332f272cbf58b4e20ae98ad6ed59a3ed1d6763R24). Since v13 it's highly recommended to use the PyPI project names in external dependencies.

joao-p-marques:
Ok, thanks @sbidoul!

joao-p-marques:
@pedrobaeza already applied the discussed changes in https://github.com/OCA/intrastat-extrastat/pull/116#discussion_r569735555

pedrobaeza (approved):

pedrobaeza:
Let's move with this:

/ocabot merge nobump

PR #124, opened 2021-02-18
[14.0][FIX] intrastat_product: Missed changes in migration

joao-p-marques:
Change back from `invoice.type` to `invoice.move_type`
Update method signature and calls after the refactor in "intrastat.product.declaration"
Add some docs to explain the functionality that leads to this part of the code.

@Tecnativa
TT28080

ping @pedrobaeza @victoralmau

### DIFF
--- a/intrastat_product/models/account_move.py
+++ b/intrastat_product/models/account_move.py
@@ -65,7 +65,12 @@ class AccountMove(models.Model):
         """
         self.mapped("intrastat_line_ids").unlink()
         for inv in self:
-            if inv.type not in ("out_invoice", "out_refund", "in_invoice", "in_refund"):
+            if inv.move_type not in (
+                "out_invoice",
+                "out_refund",
+                "in_invoice",
+                "in_refund",
+            ):
                 continue
             line_vals = []
             for line in inv.invoice_line_ids:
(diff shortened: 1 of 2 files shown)


intrastat_product/models/account_move.py
```diff
@@ -77,12 +82,16 @@ def compute_intrastat_lines(self):
...
+            "note": "",
+            "line_nbr": 0,
+        }
         decl_model = self.env["intrastat.product.declaration"]
         if decl_model._is_product(line):
             hs_code = line.product_id.get_hs_code_recursively()
             if not hs_code:
                 return vals
-            weight, qty = decl_model._get_weight_and_supplunits(line, hs_code)
+            weight, qty = decl_model._get_weight_and_supplunits(line, hs_code, notedict)
```
victoralmau, line 94: IMO isn't necessary because in `intrastat_product` addon only have 2 params.
https://github.com/OCA/intrastat-extrastat/blob/13.0/intrastat_product/models/intrastat_product_declaration.py#L268

joao-p-marques: Well, that is the 13.0 version :sweat_smile: 
This is the current method: https://github.com/OCA/intrastat-extrastat/blob/14.0/intrastat_product/models/intrastat_product_declaration.py#L327

pedrobaeza (approved):

victoralmau (approved):

PR #127, opened 2021-03-25
[14.0] mig product harmonized system delivery from Akretion

sla-subteno-it:


alexis-via:
This migration was already done in PR https://github.com/OCA/intrastat-extrastat/pull/109.

PR #129, opened 2021-04-16
[MIG] hs_code_link

hbrunn:


StefanRijnhart:
Surprised you did not use a separate commit to set the field's `store` attribute to `False`. It would be good to have that commit clarificating why setting it to `True` breaks multicompany setups.

hbrunn:
do you insist on the extra commit or wouldn't a comment in the code be more helpful?

StefanRijnhart:
A commit would be good, cause then we can cherry-pick that onto later versions!

hbrunn:
now as extra commit with comment

alexis-via:
This module is a duplicate of module product_harmonized_system_delivery which is already present in v12.0, cf https://github.com/OCA/intrastat-extrastat/tree/12.0/product_harmonized_system_delivery
So I guess we should not have 2 modules that do the same thing.

hbrunn:
thanks for the info, I've missed this module indeed. What we have here as extra feature is that we add a related field to the linked code. Should I add a PR to add this to `product_harmonized_system_delivery`?

pedrobaeza:
Isn't the twin module https://github.com/OCA/intrastat-extrastat/tree/12.0/product_harmonized_system instead of https://github.com/OCA/intrastat-extrastat/tree/12.0/product_harmonized_system_delivery?

hbrunn:
not really, those two are about handling the hs_code field the delivery module adds

alexis-via:
Yes, we can add the related field in product_harmonized_system_delivery

StefanRijnhart:
@alexis-via good suggestion! As you are here, the 13.0 version of this (apparently, somewhat redundant) module made this a stored computed field: https://github.com/OCA/intrastat-extrastat/blob/13.0/hs_code_link/models/product.py#L10. But the origin field is a company-dependent field. If you think about it, given that it's the *harmonized* system (which I interpret as, harmonized across localizations), what is the reason that you made this a company-dependent field? In one of our setups, with a handful of companies spread across the northern hafl of the globe, it does not seem to make much sense. Maybe we can change it in future editions?

hbrunn:
superseded by https://github.com/OCA/intrastat-extrastat/pull/133

PR #130, opened 2021-05-05
[14.0][FIX] intrastat_product: don't force wrong domain in moves

joao-p-marques:
The intrastat transaction domain is already forced in the model and in the field in account.move.
It was being incorrectly forced in the move, where it needed to be associated to the same company as the record, but those types can have no company associated

@Tecnativa
TT29582

ping @pedrobaeza @victoralmau

### DIFF
--- a/intrastat_product/models/account_move.py
+++ b/intrastat_product/models/account_move.py
@@ -16,6 +16,7 @@ class AccountMove(models.Model):
         tracking=True,
         check_company=True,
         help="Intrastat nature of transaction",
+        domain="['|', ('company_id', '=', False), ('company_id', '=', company_id)]",
     )
     intrastat_transport_id = fields.Many2one(
         comodel_name="intrastat.transport_mode",
--- a/intrastat_product/views/account_move.xml
+++ b/intrastat_product/views/account_move.xml
@@ -9,10 +9,7 @@
                 expr="//page[@name='other_info']//field[@name='invoice_incoterm_id']"
                 position="after"
             >
-                <field
-                    name="intrastat_transaction_id"
-                    domain="[('company_id', '=', company_id)]"
-                />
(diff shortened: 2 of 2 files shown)


victoralmau (approved):

alexis-via:
There is no domain on the field definition, cf https://github.com/OCA/intrastat-extrastat/blob/14.0/intrastat_product/models/account_move.py#L12
It's true that intrastat.transaction can have company_id = False. So the domain should be ['|', ('company_id', '=', False), ('company_id', '=', company_id)]. But I don't think it's a good idea to remove the domain.

pedrobaeza:
Put the suggested domain, but on the field, the same as in:

https://github.com/odoo/odoo/blob/44e05cd802593f6d18dcdec41fe821e30d7ae893/addons/account/models/account_move.py#L211

joao-p-marques:
Thanks @pedrobaeza @alexis-via 

Changes made :+1:

pedrobaeza (approved):

alexis-via (approved):

intrastat_product/models/account_move.py
```diff
@@ -16,6 +16,7 @@ class AccountMove(models.Model):
         tracking=True,
         check_company=True,
         help="Intrastat nature of transaction",
+        domain="['|', ('company_id', '=', False), ('company_id', '=', company_id)]",
```
rousseldenis, line 19: @joao-p-marques @pedrobaeza Shouldn't this be :

`domain="['|', ('company_id', '=', False), ('company_id', 'in', company_ids)]"`

pedrobaeza: This is not a record rule, but a domain to restrict the selectable values, and they should be according the company of the move. There's no sense on selecting something of another of the available companies for a move of a specific company.

PR #131, opened 2021-05-09
[11.0][FIX] product_harmonized_system: add migration scripts

MiquelRForgeFlow:
To avoid `local_code_company_uniq` constraint error.

It was a new constraint added in v11 due to company_id being required, and if you have some duplicated hs codes, the constraint is breaking.


MiquelRForgeFlow:
Error on CI is unrelated.

PR #132, opened 2021-05-10
[13.0][IMP] intrastat_product: add support for Brexit + small performance i…

AdriaGForgeFlow:
…mprovement + remove universal hs_code constaint

Add support for post-Brexit VAT numbers with latest python-stdnum version
Remove use of country_id.intrastat
Add 'vat' field on computation and declaration lines with a simple validity check
Add 'partner_id' field on computation lines
Prevent back to draft when an XML export is present
Remove the universal constaint applied to the hs_codes for countries in the "Europe" list, as discussed in https://github.com/OCA/intrastat-extrastat/pull/116#discussion_r569735555

Co-authored-by: Alexis de Lattre <alexis.delattre@akretion.com>

Backport of: https://github.com/OCA/intrastat-extrastat/pull/120/commits/2646dc489ed76000d5eb0356ded1a71d60bf5a67

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -7,6 +7,7 @@ import logging
 from datetime import date, datetime
 
 from dateutil.relativedelta import relativedelta
+from stdnum.vatin import is_valid
 
 from odoo import _, api, fields, models
 from odoo.exceptions import RedirectWarning, UserError, ValidationError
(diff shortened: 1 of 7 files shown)


JordiMForgeFlow:
```
def _prepare_invoice_domain(self):
        domain = [
            ("date", ">=", start_date),
            ("date", "<=", end_date),
            ("state", "=", "posted"),
            ("intrastat_country", "=", True),
            ("company_id", "=", self.company_id.id),
            ("type", "!=", "entry"),
        ]
        return domain
```


In this method, the ("intrastat_country", "=", True) should be removed

intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -424,7 +481,40 @@ def _get_incoterm(self, inv_line):
     def _get_product_origin_country(self, inv_line):
         return inv_line.product_id.origin_country_id
 
-    def _update_computation_line_vals(self, inv_line, line_vals):
+    def _get_vat(self, inv_line, notedict):
+        vat = False
+        inv = inv_line.move_id
+        if self.declaration_type == "dispatches":
```
JordiMForgeFlow:
```suggestion
        if self.type == "dispatches":
```

intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -632,14 +711,15 @@ def _gather_invoices(self):
                     "transaction_id": intrastat_transaction.id,
                     "product_origin_country_id": product_origin_country.id or False,
                     "region_id": region and region.id or False,
+                    "vat": vat,
                 }
 
                 # extended declaration
                 if self._extended:
```
JordiMForgeFlow:
```suggestion
                if self.reporting_level == "extended":
```

JordiMForgeFlow:
Making the check of the VAT a constrain raising a validation error is causing that the generation of the lines is not performed when one (or more) VAT are wrong.
```
    @api.constrains("vat")
    def _check_vat(self):
        for this in self:
            if this.vat and not is_valid(this.vat):
                raise ValidationError(_("The VAT number '%s' is invalid.") % this.vat)
```

My suggestion is to make the check in the method _gather_invoices when the VAT is taken from the line, and if there is an error specify it as a line_note like other errors.

intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -239,18 +240,74 @@ def _account_config_warning(self, msg):
...
 
-    def _get_partner_country(self, inv_line):
-        country = (
-            inv_line.move_id.src_dest_country_id
-            or inv_line.move_id.partner_id.country_id
-        )
-        if not country.intrastat:
-            country = False
-        elif country == self.company_id.country_id:
-            country = False
```
JordiMForgeFlow, line 250: This check is missing in the new changes, so we are generating export/import lines related to the company's own country.

Issue #126, opened 2021-03-08
Field hs_code_id on products and company_dependent


alexis-via:
Currently, in the module product_harmonized_system, the field hs_code_id on product.template and on product.category has company_dependent=True, cf
https://github.com/OCA/intrastat-extrastat/blob/14.0/product_harmonized_system/models/product_template.py#L16

Is it really a good idea ? Sometimes I think that it has been introduced by someone who needed it on a project in a very special context, but it adds extra complexity for everybody. In a standard scenario, HS code should not depend on company. HS code should be considered as "physical" property of a product, like the weight, which doesn't depend on the company.

What do you think ? Should be consider to switch back to company_dependent=False on v15 ?

pedrobaeza:
Yes, I think the nature of the product shouldn't changed between companies. We can even consider it for v14, but doing a migration script for "materializing" the value correctly in new physical column. This migration script will be needed anyway when switching.

azucenatrey:
Hello, we just detected the problem in 12.0 and we are going to make a modification to fix it.
When we have it, we will propose a PR to solve the issue.

PR #47, opened 2018-08-20
11.0 mig intrastat product

rvalyi:
supersedes https://github.com/OCA/intrastat/pull/30
and

- rebase on 11.0 so it includes the intrastat_base module
- cherry pick view fix from v10: https://github.com/OCA/intrastat/pull/42
- cherry pick get_hs_code_recursively on product.template form v10: https://github.com/OCA/intrastat/pull/46

However tests won't pass here yet as it depends on the product_harmonized_system to be merged here:
https://github.com/OCA/intrastat/pull/38

PR #136, opened 2021-05-19
[13.0] product_harmonized_system: do not require company on hs_code model

AdriaGForgeFlow:
This PR intends to remove the requirement of setting a company to an HS Code since HS code should be considered as "physical" property of a product which should not depend on the company and can be shared across companies.

cc ~ @ForgeFlow

### DIFF
--- a/product_harmonized_system/models/hs_code.py
+++ b/product_harmonized_system/models/hs_code.py
@@ -34,8 +34,6 @@ class HSCode(models.Model):
     company_id = fields.Many2one(
         "res.company",
         string="Company",
-        readonly=True,
-        required=True,
         default=lambda self: self._default_company_id(),
     )
     product_categ_ids = fields.One2many(


JordiMForgeFlow (approved):

StefanRijnhart:
Exactly! @alexis-via, why was this introduced in https://github.com/OCA/intrastat-extrastat/pull/47/commits/4dbfeecd09a117513efd366a221d883c31107706#diff-74e1197466e3c9a2c08b006d373e86cca69ac389f34c9e4ed38dc24b84018d80R82?

But also, why is `hs_code_id` on `product.template` a company dependent field? (https://github.com/OCA/intrastat-extrastat/blob/13.0/product_harmonized_system/models/product_template.py#L16). It looks to me that the only thing that should be company dependent is `local_code` which could be an optional field on top of a stored `hs_code` field.

astirpe (approved):
I agree with this PR. Should be backported to V11 and V12

alexis-via:
@StefanRijnhart You're not pointing to the right commit ; the diff you're pointing at is required=True on intrastat.product.declaration, which is obvisouly a good thing.

alexis-via (approved):

StefanRijnhart:
@alexis-via apologies, it was https://github.com/OCA/intrastat-extrastat/commit/981ecf85#diff-ce6878dd561fd41665ff42ec76a5b427470893007d3342eb9ab5d42fbd50b9d6R34 by @luc-demeyer.

I would still be interested to hear your thoughts about my second comment: what is the reason that `hs_code_id` on `product.template` a company dependent field? (https://github.com/OCA/intrastat-extrastat/blob/13.0/product_harmonized_system/models/product_template.py#L16). Is it not more logical to make `hs.code`'s `local_code` field company dependent instead (and make hs_code a regular column)?

luc-demeyer:
@StefanRijnhart 
I agree, local_code should be company dependent (and shareable between companies.

alexis-via:
@StefanRijnhart I agree that hs_code_id on product.template should not be company-dependant. This was made by one of my collegue who needed that on a big and complex deployment, but I was not in favor of this added-complexity. It's not always easy to be the one who go against complexity ! In fact, I created an issue 3 months ago to propose to change that, cf https://github.com/OCA/intrastat-extrastat/issues/126 I proposed to make this change in v15, but we can also make this change in v14 if we test the migration script carefully.

StefanRijnhart:
@alexis-via excellent, thanks for the pointer!

PR #138, opened 2021-07-29
[13.0][FIX] intrastat_product migration script

acsonefho:
Fix migration script who deals with `account_invoice` and `account_invoice_line` tables who don't exist in v13.0.
Add check to ensure these tables exist really (ex: project from previous version migrated to v13.0) before applying the script.

This migration script has been added in https://github.com/OCA/intrastat-extrastat/pull/111

### DIFF
--- a/intrastat_product/migrations/13.0.1.0.3/post-migration.py
+++ b/intrastat_product/migrations/13.0.1.0.3/post-migration.py
@@ -29,28 +29,31 @@ def map_intrastat_product_declaration_month(env):
 
 
 def update_invoice_relation_fields(env):
-    openupgrade.logged_query(
-        env.cr,
-        """
-        UPDATE account_move am
-        SET (intrastat_transaction_id, intrastat_transport_id,
-            src_dest_country_id, intrastat_country, src_dest_region_id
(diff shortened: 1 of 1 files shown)


intrastat_product/migrations/13.0.1.0.3/post-migration.py
```diff
@@ -29,28 +29,31 @@ def map_intrastat_product_declaration_month(env):
...
-        """
-        UPDATE intrastat_product_computation_line ipcl
-        SET invoice_line_id = aml.id
-        FROM account_invoice_line ail
-        JOIN account_move_line aml ON aml.old_invoice_line_id = ail.id
-        WHERE ipcl.%(old_line_id)s = ail.id"""
-        % {"old_line_id": openupgrade.get_legacy_name("invoice_line_id")},
-    )
+    # In a Odoo project started from v13.0, these invoices tables doesn't exist!
+    if openupgrade.table_exists(env.cr, "account_invoice"):
```
StefanRijnhart, line 33: You can minimize the diff if you just return after checking that the table does _not_ exist.

StefanRijnhart (approved):

PR #141, opened 2021-08-18
[13.0][FIX]intrastat_product_declaration - fix amount by removing leg…

luc-demeyer:
…acy currency conversion

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -344,16 +344,6 @@ class IntrastatProductDeclaration(models.Model):
 
         return weight, suppl_unit_qty
 
-    def _get_amount(self, inv_line):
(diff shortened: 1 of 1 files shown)


luc-demeyer:
we are still using a conversion from invoice currency to company currency in the intrastat product declaration,
cf. https://github.com/OCA/intrastat-extrastat/blob/02542b8fd19f6eb4d2ae1b595a6aec89cdab17ff/intrastat_product/models/intrastat_product_declaration.py#L349

This code makes no longer sense since as from Odoo 13.0 the invoice line amount in company currency is already stored in the balance field.
The currency conversion also results in deviations between the amount reported by the tax declaration and the amount reported via the intrastat declaration due to the rounding errors.
This deviation becomes considerable when having a large number of product items.
I made this fix because a customer remarked a deviation of about 200 EUR on an invoice line with quantity of about  20000.

PR #142, opened 2021-09-22
[13.0][IMP] intrastat_product: Add active field to transactions.

victoralmau:
Add `active` field to transactions.

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT30682

### DIFF
--- a/intrastat_product/models/intrastat_transaction.py
+++ b/intrastat_product/models/intrastat_transaction.py
@@ -26,6 +26,7 @@ class IntrastatTransaction(models.Model):
         string="Company",
         default=lambda self: self._default_company_id(),
     )
+    active = fields.Boolean(default=True)
 
     @api.model
     def _default_company_id(self):
--- a/intrastat_product/views/intrastat_transaction.xml
+++ b/intrastat_product/views/intrastat_transaction.xml
@@ -17,6 +17,7 @@
                     <field name="code" />
                     <field name="description" />
                     <field name="company_id" groups="base.group_multi_company" />
+                    <field name="active" />
                 </group>
             </form>
         </field>


joao-p-marques (approved):

pedrobaeza:
Several thoughts:

- Shouldn't official data should be included here, not on localization ones? I think the information and codes are standardized across whole Europe, so not sure why we should provide them repeated. The list we are handling is https://www.cbs.nl/en-gb/participants-survey/overzicht/businesses/onderzoek/international-trade-in-goods/idep-code-lists, specifically for transactions is https://www.cbs.nl/-/media/cbsvooruwbedrijf/international-trade-in-goods/code-list-13-transaction-new-code-list-2-digits-2021.pdf
- The active=False error handling may be added here, not on localization modules.

cc @alexis-via @luc-demeyer

luc-demeyer (approved):

pedrobaeza:
@luc-demeyer what do you think about my comments?

luc-demeyer:
@pedrobaeza 
I checked the list of transaction codes that we need to use in Belgium as from jan 2022 and they correspond completely to the documents you refer to, hence at least for the Belgian part. These codes are also the same in France. You are referring to a dutch document with the same codes, hence based upon same codes in at least Spain, Netherlands, France and Belgium : I agree that we should move the transaction codes to the intrastat_product module.

pedrobaeza:
@alexis-via can you pronounce about what we want to achieve. We will push it in the coming days.

rousseldenis:
> I agree that we should move the transaction codes to the intrastat_product module.

I agree too. Is there a webservice to retrieve them at once or ... ?

Shouldn't they deserve a separate module ?

luc-demeyer:
@rousseldenis 
>>> I agree too. Is there a webservice to retrieve them at once or ... ?

You can retrieve the codes in 4 languages (EN, FR, DE, NL) from https://www.nbb.be/doc/dd/onegate/data/new_natures_of_transaction_2022_en.pdf

>>> Shouldn't they deserve a separate module ?
Seen the limited number of codes: I don't think we should have a seperate module.

alexis-via:
I had a look at the document "Code list 13 - Transaction", and its quite similar to the "code nature de la transaction", cf the specs for France (in French):
https://www.douane.gouv.fr/sites/default/files/bod/src/dana/da/Statistiques%20et%20etudes%20du%20commerce%20exterieur_21-003.pdf
page 94 (appendix 3)
But there are **several differences**:
- France has 7 additional codes compared to NL: 13, 14, 19, 29, 30, 65, 70
- NL has 6 additional codes compared to FR: 31, 32, 33, 34, 60, 71

In addition, in France, we not only have "code nature de la transaction", but we also have "code regime", which is listed on page 34+35+36 of the PDF doc for France. And, in France, "code regime" is more important than "code nature de la transaction". In fact, I store code régime in the native "code" field of intrastat.transaction, and I have an additional field "fr_transaction_code" where I store "code nature de la transaction", cf https://github.com/OCA/l10n-france/blob/14.0/l10n_fr_intrastat_product/data/intrastat_transaction.xml 

So, in the end, I don't think it's a good idea to put the transaction code in the generic module.

pedrobaeza:
OK, thanks for the answer. Let's merge then at least this active flag and let each localization in their data set, although there's some similar things. Maybe we can put these links in the main README of the repository as reference or in a pinned issue.

/ocabot merge minor

Please fw-port it to v14

alexis-via:
I just received a new document dated October 18 2021 for France:
https://www.douane.gouv.fr/sites/default/files/2021-10/18/Evolution-des-modalites-declaratives-DEB-1-janvier-2022.pdf
related to the EBS reform (European Business Statistics) which will start on January 2022 and they updated the list of "code nature de la transaction". The new list is on the last page of the PDF. It is the same list as the one for Netherlands that was given earlier in this thread except for code 6x (but 6x is a bit special anyway "Particular transactions recorded for national
purposes, such as repairs"). So, with this EBS reform starting January 2022, I'm ok if we put the new code list in the base module.

luc-demeyer:
@alexis-via 
Thanks for double-checking this for France.
@rousseldenis 
You can go ahead and include the transaction codes in the intrastat_product module.

PR #149, opened 2021-10-23
[14.0][IMP] intrastat_*: Improve tests and code

rousseldenis:
- [x] No breaking changes
- [x] Rewrite tests to help test extensions
- [x] Move common code. Some code was called in intrastat_base making reference to inexistent fields in Abstract (company_id, company_country_code, state, year, month).


rousseldenis:
I think this first one is ready.

I would say it deserves one more amount of work to increase coverage, but with this, I would tend to say you can extend it and test your modifications in as simpler way.

@luc-demeyer @alexis-via @OCA/accounting-maintainers

intrastat_product/tests/common.py
```diff
@@ -0,0 +1,113 @@
...
+        vals = {
+            "name": "Intrastat Fiscal Position",
+            "intrastat": True,
+        }
+        cls.position = cls.position_obj.create(vals)
+
+    @classmethod
+    def _init_regions(cls):
+        # Create Belgium
+        cls._create_region()
```
luc-demeyer, line 43: In Belgium the instrastat regions are Flemish Region, Walloon Region, Brussels-Capital Region.
The region field should not be used for the country

intrastat_product/tests/common.py
```diff
@@ -0,0 +1,113 @@
...
+    @classmethod
+    def _create_region(cls, vals=None):
+        values = {
+            "code": "BE",
+            "country_id": cls.env.ref("base.be").id,
+            "company_id": cls.env.company.id,
+            "description": "Belgium",
+            "name": "Belgium",
+        }
+        if vals is not None:
```
luc-demeyer, line 84: In Belgium the instrastat regions are Flemish Region, Walloon Region, Brussels-Capital Region.
The region field should not be used for the country

luc-demeyer (changes requested):

luc-demeyer:
@rousseldenis 
Great contribution, thanks a lot.
Based upon code review: only one small remark: correct tests for regions.
I think we should merge this ASAP.

@alexis-via 
Ths PR looks great to me but I think you should test your DES declaration since a lot code has been moved by Denis from the product declaration to the intrastat common class, hence you can drop a lot of code from both your DEB and DES declaration but it seems recommended to to check if nothing breaks and eventually fix in the DES before merging this PR.

intrastat_product/tests/common.py
rousseldenis: Here are only for tests purpose. But if you'd prefer I will create a real region instead

luc-demeyer: I prefer using real regions: less confusing and better coverage.

rousseldenis:
@luc-demeyer Done

luc-demeyer (approved):

rousseldenis:
@luc-demeyer Can we merge this as I will need to add a breaking change in a new PR ?

luc-demeyer:
@alexis-via 
please give green ligth to merge this one.

alexis-via:
I would have preferred to drop the class intrastat.common altogether. The synergies bought by intrastat.common were pretty low. Droppping intrastat.common would have simplified the code.

Unfortunately, this PR doesn't tackle the 2 priorities:
- migration of hs_code_id on product.template/product.category to company_dependent=False
- rewrite the warnings/log system of intrastat.product.declaration to avoid have the same warning multiple times (where the same product with missing data is present in several invoice lines for example)

luc-demeyer:
@alexis-via 

This PR adds a nice set of unit tests, hence a great improvement.

Also for me dropping intrastat_common makes sense (the belgian 'services' declaration doesn't depend on this one).
This can be done in 15.0 migration (or a seperate PR but will break your french DES declaration).

Migration of hs_code_id on product.template/product.category to company_dependent=False : that issue will soon be solved by your https://github.com/OCA/intrastat-extrastat/pull/156

Rewrite the warnings/log system: I don't have too many complaints about this one in my customer base, hence low prio for me, but improving this is imho also not a showstopper for merging this PR.

alexis-via:
Sorry if my message was not very positive. The great thing with this PR is the addition of test.
What do you think about my proposal to drop the class intrastat.common ? If you agree on that, I can work on a PR to update this PR.

rousseldenis:
> Sorry if my message was not very positive. The great thing with this PR is the addition of test. What do you think about my proposal to drop the class intrastat.common ? If you agree on that, I can work on a PR to update this PR.

This PR was not aimed to enhance code but to introduce tests.

My opinion is that we can merge this fast and then doing improvements. To avoid a too great PR, which is difficult to review

rousseldenis:
> I would have preferred to drop the class intrastat.common altogether. The synergies bought by intrastat.common were pretty low. Droppping intrastat.common would have simplified the code.
> 
> Unfortunately, this PR doesn't tackle the 2 priorities:
> 
> * migration of hs_code_id on product.template/product.category to company_dependent=False
> * rewrite the warnings/log system of intrastat.product.declaration to avoid have the same warning multiple times (where the same product with missing data is present in several invoice lines for example)

Indeed, this PR highlighted the fact that common class hasn't real added value.

I would tend to say, let's do two new PR's (based on this - for tests) with one change at a time.

luc-demeyer:
I am in favour of proposal from @rousseldenis : merge this one and do the other changes via add-on PRs.

alexis-via:
Don't worry, I'll do it. I'm the most impacted by this change because this PR breaks l10n_fr_intrastat_service, so I have to invest time anyway to repair the breakage.

alexis-via:
My PR will also fix the fact that the file "account_fiscal_position_template.py " is not imported in intrastat_base.

alexis-via:
My PR to update and fix this PR: https://github.com/acsone/intrastat-extrastat/pull/1
The main change is the removal of intrastat.common.

alexis-via:
I prefer to do all the work in a single PR and not in several PRs, because of the impact on the modules of OCA/l10n-france. It would be too time consuming to make several PRs and, every time, update the code and do the coordination in localization modules.

alexis-via (approved):

PR #150, opened 2021-10-24
[14.0][ADD] intrastat_product_lot_country_of_origin

rousseldenis:
Depends on:

- [ ] #149


intrastat_product_lot_country_of_origin/models/intrastat_product_declaration.py
```diff
@@ -0,0 +1,20 @@
...
+from odoo import models
+from odoo.fields import first
+
+
+class IntrastatProductDeclaration(models.Model):
+    _inherit = "intrastat.product.declaration"
+
+    def _get_product_origin_country(self, inv_line, notedict):
+        if inv_line.purchase_line_id and inv_line.purchase_line_id.move_ids:
+            return first(
```
luc-demeyer: Shouldn't we return an empty country of origin in case there are multiple in stead of the first one ?

luc-demeyer:
@rousseldenis 
Based upon quick code review : 
Once https://github.com/OCA/stock-logistics-warehouse/pull/1282 gets merged than it is logical that we retrieve the country of origin from the lot.

rousseldenis:
> @rousseldenis Based upon quick code review : Once [OCA/stock-logistics-warehouse#1282](https://github.com/OCA/stock-logistics-warehouse/pull/1282) gets merged than it is logical that we retrieve the country of origin from the lot.

Yes. But I need to fix this before as we need to improve code for grouping declaration lines per lot (we can have several for the same product). I wanted to do a generic method in order to override it when needed.

luc-demeyer:
@rousseldenis 
I don't understand why you need to group per lot.
The grouping of lines is a legal requirement. I am not aware that in one of the EU member states grouping per lot is required.

rousseldenis:
> @rousseldenis I don't understand why you need to group per lot. The grouping of lines is a legal requirement. I am not aware that in one of the EU member states grouping per lot is required.

https://github.com/OCA/intrastat-extrastat/pull/150

If you have one reception with two country of origin for the same product, there is only one invoice line.

But in the report you need to split. But after having looked into code, the modifications to do that are too big. We'll do it in a further version maybe.

luc-demeyer:
@rousseldenis 
Requirement clear now. I think though that this use case is too specific to include in a generic OCA module. A small custom module for your customer project is imho the best way forward.

rousseldenis:
> @rousseldenis Requirement clear now. I think though that this use case is too specific to include in a generic OCA module. A small custom module for your customer project is imho the best way forward.

Yes and no as it implies great refactor on how lines are retreived. But we pause this dev for now.

But I will think about it on how to add a minimum of hooks in standard module to do it.

PR #156, opened 2021-12-13
[14.0][MIG] hs_code_id company_dependent=True to False

alexis-via:
Migration script provided for product.template and product.category
Fixes bug #126

### DIFF
--- a/product_harmonized_system/migrations/14.0.2.0.0/post-migration.py
+++ b/product_harmonized_system/migrations/14.0.2.0.0/post-migration.py
@@ -1,0 +1,45 @@
+# Copyright 2021 Akretion France (http://www.akretion.com/)
+# @author: Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import SUPERUSER_ID, api
+
+
+def migrate(cr, version):
+    if not version:
+        return
+
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        pc_field_id = env.ref(
+            "product_harmonized_system.field_product_category__hs_code_id"
+        ).id
+        cr.execute(
+            """
+            UPDATE product_category pc
+            SET hs_code_id=SUBSTRING(ip.value_reference, 9, 99)::int
+            FROM ir_property ip
+            WHERE ip.res_id like 'product.category,%%' AND
+            SUBSTRING(ip.res_id, 18, 99)::int=pc.id AND
+            ip.name='hs_code_id' AND
+            ip.value_reference IS NOT null AND
+            ip.fields_id=%s
+            """,
+            (pc_field_id,),
+        )
+        pt_field_id = env.ref(
+            "product_harmonized_system.field_product_template__hs_code_id"
+        ).id
+        cr.execute(
+            """
+            UPDATE product_template pt
+            SET hs_code_id=SUBSTRING(ip.value_reference, 9, 99)::int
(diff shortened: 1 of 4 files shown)


alexis-via:
Note: we have to merge this BEFORE migration to v15 starts #144

pedrobaeza (changes requested):
I think you should do an SQL query on this form instead of such complicated code that is going to be very slow:

```sql
UPDATE product_category pc
SET hs_code_id = SUBSTRING(ip.value_reference, 7, 99)::int
FROM ir_propery ip
WHERE ip.res_id = pc.id AND ip.name = 'hs_code_id'
```
The number 7 is the position for starting to extract the number, and I'm not sure if it's the right one, but I think you get the point.

alexis-via:
SQL 1, ORM 0

product_harmonized_system/migrations/14.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,49 @@
...
+        ).id
+        cr.execute(
+            """
+            UPDATE product_category pc
+            SET hs_code_id=SUBSTRING(ip.value_reference, 9, 99)::int
+            FROM ir_property ip
+            WHERE ip.res_id like 'product.category,%%' AND
+            SUBSTRING(ip.res_id, 18, 99)::int=pc.id AND
+            ip.name='hs_code_id' AND
+            ip.company_id IS NOT null AND
```
pedrobaeza: I don't think you should restrict this. Maybe there's a generic value for all companies in the DB. ir.property allows it...

product_harmonized_system/migrations/14.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,49 @@
...
+            ip.value_reference IS NOT null AND
+            ip.type='many2one' AND
```
pedrobaeza: If you are already matching by the field id, I think this is too much, but I don't think this is going to increase too much the query time.

pedrobaeza:
> SQL 1, ORM 0

Welcome to the usual way on OpenUpgrade!

product_harmonized_system/migrations/14.0.2.0.0/post-migration.py
alexis-via: done

product_harmonized_system/migrations/14.0.2.0.0/post-migration.py
alexis-via: done

pedrobaeza (approved):

sebalix (approved):

product_harmonized_system/migrations/14.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,45 @@
...
+        pt_field_id = env.ref(
+            "product_harmonized_system.field_product_template__hs_code_id"
+        ).id
+        cr.execute(
+            """
+            UPDATE product_template pt
+            SET hs_code_id=SUBSTRING(ip.value_reference, 9, 99)::int
+            FROM ir_property ip
+            WHERE ip.res_id like 'product.template,%%' AND
+            SUBSTRING(ip.res_id, 18, 99)::int=pt.id AND
```
hparfr, line 39: Shound't be 17 instead ?

`SUBSTRING(ip.res_id, LENGTH('product.product,') +1 )::int=pt.id AND`

hparfr: my bad, was reviewing the wrong field !

PR #160, opened 2021-12-26
[13.0]Intrastat Brexit support

luc-demeyer:
This PR now handles correctly the XI and XU country codes introduced by the brexit.


luc-demeyer:
@alexis-via 
I kept the now depecrrated product_origin_country_id field to avoid breaking the localisation modules.
I have tested the change with  l10n_fr_intrastat_product module.

PR #162, opened 2021-12-31
[13.0]intrastat_product - fix amount_company_currency

luc-demeyer:


luc-demeyer:
This PR fixes incorrect calculation for incoming transactions and refunds (error introduced by https://github.com/OCA/intrastat-extrastat/pull/141).

PR #163, opened 2022-01-02
[14.0]intrastat Brexit support

luc-demeyer:


pedrobaeza:
Please rebase/cherry-pick instead of pulling for avoiding the merge commits.

luc-demeyer:
@alexis-via 
I have tested the changes in this module with your l10n_fr_intrastat_product localisation and it looks ok: my changes seem to be non-breaking for the localisation modules but imho we should forward port this PR asap to 15.0 and remove the deprecated fields and logic from the localisation modules.

While testing l10n_fr_intrastat_product I encountered what I believe is a functonal bug (but unrelated to the brexit change in this PR). I have created a PR for this one, cf. https://github.com/OCA/l10n-france/pull/357

I was also surprised to find the _get_product_origin_country in your module with a warning on missing product country of origin. In Belgium we can simply specify 'QU' as country code in this case.

intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -490,8 +490,30 @@ def _get_incoterm(self, inv_line, notedict):
         return incoterm
 
     def _get_product_origin_country(self, inv_line, notedict):
+        _logger.warning(
```
ghost: In Odoo Community, they are using warnings.warn with a DeprecationWarning and this warning is shown in logs but also in the IDE (at least Pycharm does it) as Deprecated https://github.com/odoo/odoo/blob/687d88eeea836aacbcd1b079e82e597d1bd9309c/addons/account/models/account_move.py#L119

intrastat_product/models/account_move.py
```diff
@@ -176,3 +197,25 @@ def _onchange_move_id(self):
...
+    def write(self, vals):
+        self._format_vals(vals)
+        return super().write(vals)
+
+    def _format_vals(self, vals):
+        if "product_origin_country_code" in vals:
+            vals["product_origin_country_code"] = (
+                vals["product_origin_country_code"].upper().strip()
+            )
+            if len(vals["product_origin_country_code"]) != 2:
```
ghost: Instead of having a check here on the length, I think it would be more easier and user-friendly to have a max-length directly on the field to be warned beforehand.

luc-demeyer:
@jdidderen-be 
Thanks for your review. Code adapted accordingly.

ghost (changes requested):

ghost (approved):

vanderperre (commented):
Hi Luc,


the version must take into account of the 14.0.1.5.4 changes ....

When i copy/paste in my current last oca update ....:

     "name": "Intrastat Product",
-    "version": "14.0.1.5.4",
+    "version": "14.0.1.5.3",
     "category": "Intrastat",

A+

Screen Shot 2022-05-12 at 12 14 59

aaah Ok, the versioning is automatic with the runbot ... ;)

vanderperre (approved):

alexis-via:
With this code, we now have this bug when a French company sells to a belgian companies but the goods are delivered to spain: cf last line of the screenshot
Src/Dest Country = Spain
Country code = BE

bug_intrastat

This is because the country code is now only generated by looking at the commercial_partner_id of the invoice, without consideration for the field "src_dest_country_id". It's great to have super code for Northern-Ireland, but I should not break basic country code scenarios!

alexis-via:
Other issues caused by this PR:
- we don't see the product country of origin any more, but only it's code, which makes it more difficult for the user to control the data before validating the declaration.
- the field "origin_state_id" is always displayed on product form view in the first tab, which is really a waste given that's it is only for products from Northern Ireland... I don't think we can afford that on the product form ! => what do you think about displaying it only when the country is UK ?

I really think that we shouldn't have gone so far for Northern Ireland. Northern Ireland is a mess anyway, so we should not try to make the "perfect" implementation with so many drawbacks for all users. I think my previous implementation was "good enough" and didn't cause so many drawback for such a corner-case scenario.

intrastat_base/models/res_partner.py
```diff
@@ -0,0 +1,67 @@
...
+    def _get_xi_counties(self):
+        return [self.env.ref(x) for x in XI_COUNTIES]
+
+    @api.model
+    def _get_xu_counties(self):
+        uk_counties = self.env.ref("base.uk").state_ids
+        xu_counties = uk_counties.filtered(lambda r: r not in self._get_xi_counties())
+        return xu_counties
+
+    def _get_intrastat_country_code(self, country=None, state=None):
```
alexis-via, line 46: This method is really strange with it's "if self / else" ! When self has a value, then the arguments country and state are ignored, which is surprising. For me, instead of this wired code, we should have 2 separate methods, the first one calling the second one.

intrastat_product/models/account_move.py
```diff
@@ -160,11 +169,23 @@ class AccountMoveIntrastatLine(models.Model):
...
     product_origin_country_id = fields.Many2one(
         comodel_name="res.country",
         string="Country of Origin",
         help="Country of origin of the product i.e. product " "'made in ____'.",
     )
+    product_origin_country_code = fields.Char(
+        string="Country of Origin of the Product",
+        size=2,
+        required=True,
+        default="QU",
```
alexis-via, line 184: I find it a bit "too easy" to use QU by default. I think we should display a warning in the warning dialog box about the fact that the country of origin is missing on product X and say that we've set the country code to QU as a temporary solution.

intrastat_product/views/intrastat_product_declaration.xml
```diff
@@ -287,11 +294,17 @@
...
+                <!-- TODO: remove product_origin_country_id after update of localization modules -->
                 <field
                     name="product_origin_country_id"
                     invisible="1"
                     string="Product C/O"
                 />
+                <field
+                    name="product_origin_country_code"
+                    attrs="{'column_invisible': [('parent.declaration_type', '=', 'arrivals')]}"
+                    string="Product C/O"
```
alexis-via, line 306: In France, since for reform of January 2022 1st, we have to declare the product country of origin on arrivals AND departures! I would be very surprised if it was not the case in other countries. Cf page 28 section "9. Pays d'origine" of this PDF : https://www.douane.gouv.fr/sites/default/files/2021-12/21/Note-enquete-statistique-EMEBI-2022.pdf

PR #172, opened 2022-05-02
[14.0]intrastat_product - remove incorrect domain on comutation line

luc-demeyer:
The intrastat flag has been removed from res.country via https://github.com/OCA/intrastat-extrastat/commit/b7ba7f7f5109d6ab1a47f6b6b261187f02193ea3 but was still hanging around in the computation lines resulting in stack trace when trying to change the country manually on a computation line.


pedrobaeza (approved):

pedrobaeza:
Fast-tracking as obvious:

/ocabot merge patch

PR #176, opened 2022-06-08
[13.0][OU-ADD] product_harmonized_system: Multi-company rules migration script.

victoralmau:
Multi-company rules migration script.

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT37373


pedrobaeza (commented):
I'm thinking that this will update for everyone (coming from migration or not) to force such domain. Maybe it's better to let them at 13.0.1.0.0 and change manually already migrated databases

pedrobaeza (approved):

Issue #175, opened 2022-05-25
[RFC] intrastat_product: Add an Intrastat summary in the invoice report


pedrobaeza:
One of my customers has asked for having an Intrastat summary on each invoice for their customers to be aware about the HS codes, weights and so on to fill. I think this can be something that can be done in this module instead of handling it on each localization.

It can be an extra page in the same invoice report with a table with the following cells:

- Product reference
- Product name
- HS code
- Weight
- Origin country

And it would be something opt-in to be marked on the commercial partner.

What do you think? @alexis-via @luc-demeyer

luc-demeyer:
@pedrobaeza 
this is already there : cf. intrastat transaction details.

pedrobaeza:
But is there a report to give it to the customer?

luc-demeyer:
the info is in the invoice hence adapt the invoice qweb template and it's done

pedrobaeza:
And this RFC is for that task :wink:

PR #177, opened 2022-06-09
[13.0][IMP] intrastat_product: Add an Intrastat summary in the invoice report.

victoralmau:
Related to: https://github.com/OCA/intrastat-extrastat/issues/175

Add an Intrastat summary in the invoice report.

@Tecnativa TT37168


intrastat_product/models/account_move.py
```diff
@@ -101,6 +101,36 @@ def _get_intrastat_line_vals(self, line):
             )
         return vals
 
+    def _get_intratat_line_info(self, line):
```
pedrobaeza:
```suggestion
    def _prepare_intrastat_line_info(self, line):
```

intrastat_product/models/res_partner.py
```diff
@@ -0,0 +1,16 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ResParter(models.Model):
+    _inherit = "res.partner"
+
+    invoice_intrastat_detail = fields.Boolean(
+        string="Show intrastat detailt in invoice report"
```
pedrobaeza:
```suggestion
        string="Show Intrastat details in invoice report"
```

intrastat_product/views/report_invoice.xml
```diff
@@ -0,0 +1,52 @@
...
+        inherit_id="account.report_invoice_document"
+    >
+        <xpath expr="//div[hasclass('page')]" position="inside">
+            <t t-if="o.partner_id.invoice_intrastat_detail">
+                <t
+                    t-set="intrastat_lines_info"
+                    t-value="o._get_intrastat_lines_info()"
+                />
+                <t t-if="intrastat_lines_info">
+                    <table
```
pedrobaeza: Put a header saying "Intrastat information" or similar

victoralmau:
Changes done

pedrobaeza (commented):
If the same product is in several lines, they are not grouped.

victoralmau:
> If the same product is in several lines, they are not grouped.

Ok, I am going to group the information so that the same product is not displayed several times (by default, the module does not group the records with respect to the field "Intrastat declaration details" `intrastat_line_ids`)

victoralmau:
Changes done, now information is grouped by product.

intrastat_product/models/account_move.py
```diff
@@ -90,17 +90,64 @@ def _get_intrastat_line_vals(self, line):
...
+                if is_intrastat_line
+                else product.origin_country_id
+            ),
+        }
+
+    def _get_intrastat_lines_info(self):
+        """We obtain a list of information that we will need to group at the end by
+        product and sum weight.
+        """
+        items = (
```
pedrobaeza: Do this in a more readable way, doing a loop and using `setdefault`.

intrastat_product/models/account_move.py
```diff
@@ -90,17 +90,61 @@ def _get_intrastat_line_vals(self, line):
             if not hs_code:
                 return vals
             weight, qty = decl_model._get_weight_and_supplunits(line, hs_code)
+            if not self.env.context.get("bypass_transaction_weight_convert"):
```
pedrobaeza: Why this?

intrastat_product/models/account_move.py
pedrobaeza: This initial list shouldn't be needed

intrastat_product/models/account_move.py
```diff
@@ -90,17 +90,61 @@ def _get_intrastat_line_vals(self, line):
...
+            )
+            if not self.intrastat_line_ids
+            else self.intrastat_line_ids
+        )
+        res = {}
+        for line in items:
+            res.setdefault(line.product_id.id, {"weight": 0})
+            vals = self._prepare_intrastat_line_info(line)
+            weight = vals["weight"]
+            del vals["weight"]
```
pedrobaeza: Not sure why you need this, but in any case, it can be done as:

```suggestion
            weight = vals.pop("weight")
```

intrastat_product/models/account_move.py
```diff
@@ -90,17 +90,61 @@ def _get_intrastat_line_vals(self, line):
...
+            res[line.product_id.id].update(vals)
+            res[line.product_id.id]["weight"] += weight
+        return [res[key] for key in res]
```
pedrobaeza:
```suggestion
        return res.values()
```

intrastat_product/models/account_move.py
victoralmau: This is necessary to prevent weight being converted to an integer by default.

intrastat_product/models/account_move.py
victoralmau: Ok, but without this list how do we iterate through each element to get the final list?

pedrobaeza: Do the loop over it in the same way.

intrastat_product/models/account_move.py
pedrobaeza: Yes, but why convert it to int? Let the decimal precision to act

intrastat_product/models/account_move.py
victoralmau: Maybe I don't understand what you are saying, but it is necessary to iterate all the invoice lines (or intrastat_line_ids) so that the sum of the weight is correct.

pedrobaeza (approved):

pedrobaeza:
@luc-demeyer @alexis-via is it OK for you?

victoralmau:
Rebase done to resolve conflicts.

pedrobaeza:
Merging as no further feedback:

/ocabot merge major

PR #181, opened 2022-06-14
[13.0][IMP] intrastat_base: use fiscal position

victoralmau:
Backport from 14.0: https://github.com/OCA/intrastat-extrastat/pull/115

@Tecnativa TT37486


pedrobaeza:
Why do we need this over-complication?

victoralmau:
> Why do we need this over-complication?

This is necessary to avoid the test error https://github.com/OCA/intrastat-extrastat/runs/6879754936?check_suite_focus=true#step:8:129

Looks like this still needs to be applied: https://github.com/OCA/intrastat-extrastat/pull/120/commits/2646dc489ed76000d5eb0356ded1a71d60bf5a67

JordiBForgeFlow:
@victoralmau Now when you use the intrastat in v13 the country of the company should not have the field 'intrastat' set, right? With this PR this is irrelevant, as it is the fiscal position what drives the recognition that this is an intrastat transaction. For example, if my company is in spain, then even when Spain is in the EU, the field 'intrastat' should NOT be set.

pedrobaeza:
Check #192

pedrobaeza:
@victoralmau now that I re-read the original thread, what we need to adapt is for v14, the question of the fiscal positions having the check or not. Please check. I would remove in the localization such use case.

victoralmau:
> @victoralmau now that I re-read the original thread, what we need to adapt is for v14, the question of the fiscal positions having the check or not. Please check. I would remove in the localization such use case.

I'm sorry but I'm not sure exactly what I need to check. Is it related to this https://github.com/OCA/l10n-spain/blob/14.0/l10n_es_intrastat_report/data/account_fiscal_position_template.xml ?

pedrobaeza:
Please read #115 and the adaptation to be done is to verify the check at FP.

PR #183, opened 2022-06-22
[14.0][IMP] intrastat_product: pre_init_hook

StefanRijnhart:
On a large database, this can speed up the time to deploy this module more than 10 times.

### DIFF
--- a/intrastat_product/__init__.py
+++ b/intrastat_product/__init__.py
@@ -1,3 +1,4 @@
 from . import models
 from . import report
 from . import wizards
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):
I think you can pass any number of arguments to coalesce and it will use the first not null.

StefanRijnhart:
> I think you can pass any number of arguments to coalesce and it will use the first not null.

Thanks, fixed!

MiquelRForgeFlow (approved):

PR #189, opened 2022-08-10
[13.0][FIX] intrastat_product: Don't add notes to the variable.

victoralmau:
Don't add notes to the variable: `_format_line_note()` function adds it to the `notedict` (in some parts of the code it is correct).

This change was added at https://github.com/OCA/intrastat-extrastat/commit/446dcb3f4e7b19b240a99d7e185b9af24e12cd37 (forgot part of what was changed at https://github.com/OCA/intrastat-extrastat/commit/2646dc489ed76000d5eb0356ded1a71d60bf5a67).

In v14 it is correct.

Related to https://github.com/OCA/l10n-spain/pull/2465

@Tecnativa

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -336,7 +336,7 @@ class IntrastatProductDeclaration(models.Model):
 
         if not source_uom:
             line_notes = [_("Missing unit of measure.")]
-            self._note += self._format_line_note(inv_line, notedict, line_notes)
+            self._format_line_note(inv_line, notedict, line_notes)
             return weight, suppl_unit_qty
 
         if intrastat_unit_id:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I don't get the problem, and this is modifying the signature of one method, thus affecting potentially to other dependent modules.

victoralmau:
`_format_line_note()` function doesn't return anything, that's why the error in tests related to `l10n_es_intrastat_report`.

In `l10n_es_intrastat_report` notes are added, and `notedict` is needed to do so, so it is added to the `_update_computation_line_vals()` function.

victoralmau:
Ok, right now, this change will resolve the error in the `l10n_es_intrastat_report` tests.

pedrobaeza:
And why you don't explain this from the beginning and include such explanation in the commit message? Why the signature change?

pedrobaeza:
Please detail in the commit message the problem that is fixing. Is this happening as well in later versions?

victoralmau:
The commit message is now updated.
This is fixed in v14 (https://github.com/OCA/intrastat-extrastat/commit/2646dc489ed76000d5eb0356ded1a71d60bf5a67)

pedrobaeza (approved):

PR #192, opened 2022-09-09
[13.0][IMP] intrastat_product: Skip invoices according to partner country

victoralmau:
Skip invoices according to partner country.

When the country is the same as the company's country or the country does not have the intrastat check marked, it must be skipped for the declaration.

Accidentally removed in https://github.com/OCA/intrastat-extrastat/commit/3713df54f0230fe3059321ce399618074ef3b756

@Tecnativa TT38922

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -625,6 +625,13 @@ class IntrastatProductDeclaration(models.Model):
                 partner_country = self._get_partner_country(
                     inv_line, notedict, eu_countries
                 )
+                # When the country is the same as the company's country or the country
+                # does not have the intrastat check marked, it must be skipped.
+                if (
+                    not partner_country.intrastat
+                    or partner_country == self.company_id.country_id
+                ):
+                    continue
                 partner_country_code = (
                     invoice.commercial_partner_id._get_intrastat_country_code()
                 )


intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -625,6 +625,15 @@ def _gather_invoices(self, notedict):
                 partner_country = self._get_partner_country(
                     inv_line, notedict, eu_countries
                 )
+                if not partner_country.intrastat:
+                    line_notes = [_("Country without intrastat option checked")]
```
pedrobaeza: I don't think we should log both cases.

intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -625,6 +625,14 @@ def _gather_invoices(self, notedict):
...
                     inv_line, notedict, eu_countries
                 )
+                # When the country is the same as the company's country or the country
+                # does not have the intrastat check marked, it must be skipped.
+                if (
+                    not partner_country.intrastat
+                    or partner_country == self.company_id.country_id
+                ):
+                    continue
+
```
pedrobaeza: No extra empty lines, please

pedrobaeza (approved):

chienandalu (approved):

PR #195, opened 2022-10-14
[FIX]intrastat declaration - weight is optional for hs codes with sup…

luc-demeyer:
…plementary unit


pedrobaeza (approved):
/ocabot merge patch

Please fw-port it

PR #197, opened 2022-10-15
[FIX] intrastat declaration - restore _compute_numbers that was dropped by …

luc-demeyer:
…PR #149 (removal of intrastat.common


luc-demeyer:
No reaction after 29 days but I think it's a no brainer that this code that was dropped accidentally must be restored.

/ocabot merge patch

PR #198, opened 2022-11-04
[14.0][IMP] Big simplification, update and cleanup

alexis-via:
This commit aims at removing the over-complexity of intrastat modules while increasing simplicity/usability for users. Some of you will probably disagree with some of the change in propose in this PR. But I have a strong disagreement with the over-complexity and the over-engineering that was introduced by this Brexit PR https://github.com/OCA/intrastat-extrastat/pull/163 For me, this over-complexity has become so expensive in term of software maintenance that, if the changes that I propose here are not accepted in v16, my intrastat v16 module for France will not be based on intrastat_product any more. So far, developping the generic module "intrastat_product" was an investment to try to reach the "ideal" where the biggest part of the code possible is generic. But, when looking at the past experience and where we are today with the codebase and the experience for the user, I think it may not have been the best decision in term of user experience and cost of software maintenance.

There are so many changes that I don't really expect to have it merged in v14 (If you want to merge it in v14, I'll add the migration scripts). But it's a good basis of discussion for v16 with a concrete example. My l10n_fr_intrastat_product module for v14 will be based on this PR (it was broken following the big Brexit changes of PR 163, so I had to fix it anyway).

Here are the main changes:
- Move default intrastat transaction from res.company to account.fiscal.position to add support for B2C (and not just B2B)
- improve usability: auto-generate declaration lines and XML export when going from draft to done. Auto-delete declaration lines and XML export when going from done to draft (and add confirmation pop-up).
- declaration lines are now readonly. Only computation lines can be created/edited manually
- add field region_code on computation lines and declaration lines. Remove region_id on declaration lines. This change allows big simplification in some localization modules such as l10n_fr_intrastat_product.
- simplify Brexit implementation. Northern Ireland is important, but we can't afford to have so many lines of code and add a field on product.template (origin_state_id) for a territory of 1.9 million inhabitants! This is too costly to maintain and too complex for users.
- improve default visibility of fields when reporting_level = 'standard'
- add support for weight calculation from uom categories other than units and weight, supposing that the 'weight' field on product.template is the weight per uom of the product
- add EU companies from several different countries in demo data with valid VAT numbers


pedrobaeza:
@victoralmau please check this

intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -294,70 +285,42 @@ def _get_partner_country(self, inv_line, notedict, eu_countries):
...
-                        "its new VAT number starting with 'XI' following Brexit."
-                    )
-                    % (
-                        inv.name,
-                        inv.fiscal_position_id.display_name,
-                        vat,
-                        inv.commercial_partner_id.display_name,
-                    )
-                ]
-                self._format_line_note(inv_line, notedict, line_notes)
```
victoralmau, line 345: Can we maintain these warnings with this new approach?

intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -1131,67 +1065,66 @@ class IntrastatProductComputationLine(models.Model):
...
+        "intrastat.transaction",
+        string="Intrastat Transaction",
     )
     region_id = fields.Many2one("intrastat.region", string="Intrastat Region")
-    # product_origin_country_id is replaced by product_origin_country_code
-    # this field should be dropped once the localisation modules have been
-    # adapted accordingly
+    # Note that, in l10n_fr_intrastat_product and maybe in other localization modules
+    # region_id is left empty and Odoo writes directly in region_code
+    region_code = fields.Char()
```
victoralmau, line 1112: Perhaps another option would be to make the compute store=True field to avoid the onchange and be able to overwrite it in other modules.

victoralmau (changes requested):
First of all thank you very much for trying to simplify the logic and unnecessary fields of these modules (even if the diff is large).

I think I understand the reason for removing the fields in r`es.company` and adding them now in `account.fiscal.position`, would a migration script be necessary to define the corresponding values in all fiscal positions and thus maintain a similar behavior to the current one?
What do you think about this approach @pedrobaeza?

Could you try to separate all these changes in small commits easier to review (I know this is a bit difficult now, but it will help to understand the changes better and also in the future if it is merged)?

pedrobaeza:
Why not having both the fields in `account.fiscal.position` and `res.company` as fallback?

luc-demeyer:
We must welcome simplifications but unfortunately legal reports must also adhere to legal requirments.

I agree with some of the proposed changes but imho we should go a lot further.
We should move more logic from the localisation modules to the base modules hence making the localisation modules smaller and hence easier to maintain.

I also think we can go further in simplifying the user experience while keeping at least the current level of legal compliancy (for compliancy we should imho improve the link with the shipments, this is e.g. required in order to correctly handle refunds).

It's a very complex legal report and the legal specs are also changing constantly (as we have seen with the brexit, the harmonisation of the transactions codes, the b2b & b2c handling, ... ) and unfortunately not with the same speed in the different EU member countries.

I think it's going to be hard  (unproductive)  to take the right decisions on base module functionality versus localisation modules via a github thread.

What about organising a 3 days code sprint with a limited set of people having a good functional functional view on the regulations ? We currently have the french, spanish and belgian localisaiton modules and the generic declaration for use in the other countries. Hence I propose a code sprint with representatives of at least France (Alexis ?), Belgium (myself) and Spain where we define on how to move forward and create the 16.0 intrastat modules.
I can host it in our offices if we have a limited set of participants (<10) but I have no problem moving to France or Spain either.

A question specifically for @pedrobaeza  :
would be hard to create and 'extended' pipeline so that any change to e.g. the intrastat_product module also triggers the pipelines of the localisation modules in the l10n-belgium, l10n-france, l10n-spain, ... repos ?

luc-demeyer:
in the meanwhile I figured out that our italian colleagues also maintain an intrastat module (cf. https://github.com/OCA/l10n-italy/blob/14.0/l10n_it_intrastat) but this one does not build upon intrastat_product.
Hence I hope they are also willing to help to build a solid base module.

pedrobaeza:
It's not feasible to make inter-repositories checks from a change on the base one, sorry.

intrastat_product/models/intrastat_product_declaration.py
alexis-via: No, because, in l10n_fr_intrastat_product, I don't use the object "intrastat.region", because I already have another object for that in the french localisation. That's why I want to have the char field region_code.

alexis-via:
@pedrobaeza having the m2o fields for intrastat.transaction both on account.fiscal.position and res.company as fallback is a possibility. I didn't feel the need for it because I usually have 1 or 2 fiscal position for intrastat, not more. But, if you do B2C and you are above the 10k€/year limit, then I suppose that you end up with 26 intrastat B2C fiscal positions, and that's where it could be an advantage to have the fallback configuration on res.company.

alexis-via:
@luc-demeyer Thanks for your proposal to organise a code sprint on intrastat. But, for me, it illustrates the problem of this generic module: it is very complex and time-consuming to agree on the implementation details of this generic module. And, in all the very important things that we need to work on to improve OCA accounting (bank statement, payment orders, ...), intrastat should not be so high on the list. In other words : if I wanted to participate to an OCA code sprint in the coming months, I would prefer to have it on other topics than intrastat.

When I made this PR, I was quite upset about the situation of the intrastat code base, in particular about the over-complexity for Brexit/Northern Ireland and I was evaluating the decision to stop using intrastat_product.
I must say I was expecting more critics on this PR, especially on the removal of the file intrastat_base/model/res_partner.py and the removal of the field origin_state_id on product.template, which are the 2 main points that illustrate the over-complexity of the Brexit implementation.

So, reading the current comments on the PR, I have the feeling that you agree on my proposal. @luc-demeyer are you really ok about this, in particular the removal of intrastat_base/model/res_partner.py and the removal of the field origin_state_id on product.template ?

intrastat_product/models/intrastat_product_declaration.py
alexis-via: Yes, we could keep the second warning and restrict it for B2B (vat_required=True on fiscal position). We should drop the first warning (or make it generic for B2B when vat_required=True on the fiscal position.

alexis-via:
@pedrobaeza I implemented your idea in this PR: default intrastat transaction is now both on fiscal position and company.
Now, Odoo uses by order of priority:
1) intrastat transaction of the invoice
2) default intrastat transaction of the fiscal position of the invoice
3) default intrastat transaction of the company of the invoice
I like this solution!

alexis-via:
OK guys, I've been dreaming of doing this for so many years, here it is:
intrastat_new_logs

No more double warning messages. Nice HTML display. Warnings grouped by section and record.

pedrobaeza:
Nice. Please check CI.

alexis-via:
CI is now green

intrastat_product/models/account_move.py
```diff
@@ -55,10 +55,6 @@ def _compute_src_dest_country_id(self):
                 country = inv.company_id.country_id
             inv.src_dest_country_id = country.id
 
-    @api.model
-    def _default_src_dest_region_id(self):
```
pedrobaeza, line 59: This hook was done for a reason, so it's not good to remove it.

intrastat_product/wizards/intrastat_result_view.py
```diff
@@ -9,6 +9,6 @@ class IntrastatResultView(models.TransientModel):
     _name = "intrastat.result.view"
     _description = "Pop-up to show errors on intrastat report generation"
 
-    note = fields.Text(
+    note = fields.Html(
```
pedrobaeza, line 12: This requires a migration script to convert the plain text to html. There's an openupgradelib method for it. On pre, copy the column to preserve the old content untouched, and on post, do the conversion.

alexis-via: This data is in a wizard, so the data is not kept.

pedrobaeza: Ok then

intrastat_product/views/intrastat_region.xml
```diff
@@ -4,7 +4,7 @@
         <field name="name">intrastat.region.form</field>
         <field name="model">intrastat.region</field>
         <field name="arch" type="xml">
-            <form string="Intrastat Region">
+            <form>
```
fmdl, line 7: Don't forget to add sheet.

PR #205, opened 2022-12-12
[14.0][FIX] intrastat_product: Set the correct country code when the country of the delivery address is different (and to be consistent with the src_dest_country_id field).

victoralmau:
FWP from 13.0: https://github.com/OCA/intrastat-extrastat/pull/202

Related to https://github.com/OCA/l10n-spain/pull/2672

Set the correct country code when the country of the delivery address is different (and to be consistent with the `src_dest_country_id` field).

@Tecnativa TT40764


pedrobaeza (approved):

victoralmau:
Ping @chienandalu

chienandalu (approved):

PR #179, opened 2022-06-14
[13.0][IMP] intrastat*: brexit support

victoralmau:
Backport from 14.0: https://github.com/OCA/intrastat-extrastat/pull/163

Please @pedrobaeza can you review it?

@Tecnativa TT37486

PR #206, opened 2022-12-17
[13.0][IMP] intrastat_product - add notedict to some methods

ghost:
I add the notedict dictionary on some methods to allow in our l10n_be_intrastat_product_declaration module (https://github.com/OCA/l10n-belgium/blob/14.0/l10n_be_intrastat_product/models/l10n_be_intrastat_product_declaration.py#L48) to adapt our backport to v13 to fit with the changes made in the PR #179 (https://github.com/OCA/intrastat-extrastat/blob/13.0/intrastat_product/models/intrastat_product_declaration.py#L766)

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -31,11 +31,6 @@ class IntrastatProductDeclaration(models.Model):
             "or change the revision number of this one.",
         )
     ]
-    # TODO:
-    # drop the use of self._note & self._line_nbr when migrating to Odoo 14
-    # refactoring this now may break the localisation modules
-    _note = ""
-    _line_nbr = ""
 
     @api.model
     def default_get(self, fields_list):
@@ -308,7 +303,7 @@ class IntrastatProductDeclaration(models.Model):
                 self._format_line_note(inv_line, notedict, line_notes)
         return country
 
-    def _get_intrastat_transaction(self, inv_line):
+    def _get_intrastat_transaction(self, inv_line, notedict):
         invoice = inv_line.move_id
         if invoice.intrastat_transaction_id:
             return invoice.intrastat_transaction_id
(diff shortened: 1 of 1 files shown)


luc-demeyer (changes requested):
changing the signature of the methods will break the localisation modules.
I suggest to change it to e.g.
def _get_intrastat_transaction(self, inv_line, notedict=None)

pedrobaeza:
The same problem will happen with keyword arguments. The only solution I see now for this version is to pass the variable by context.

luc-demeyer:
@pedrobaeza 
The problem starts with https://github.com/OCA/intrastat-extrastat/pull/179 where not only the brexit but also the 'notedict' changes have been backported to 13.0 resulting in broken french and belgian localisation modules. We cannot revert it back since otherwise we'll break the spanisch 13.0 l10n_es_intrastat_report.
In order to move forward I propose following actions
1) we backport the 'notedict' based l10n_be_intrastat_product from 14.0 to 13.0
2) we do the same for l10n_fr_intrastat_product

@alexis-via 
Do you agree with this approach ? We'll backport the french module and test it.
We count on you for review and merge.

pedrobaeza:
If you change all the OCA involved modules, OK then.

luc-demeyer:
I am only aware of
-  l10n_fr_intrastat_product
-  l10n_es_intrastat_report
-  l10n_be_intrastat_product
We'll cover these.

pedrobaeza:
Ok

alexis-via:
@luc-demeyer ok. I don't use v13 myself, so it's not a big deal for me.

victoralmau (approved):
Changes seem to make sense.

pedrobaeza:
@jdidderen-be please squash both commits into one.

ghost:
> @jdidderen-be please squash both commits into one.

@pedrobaeza Commits are squashed

pedrobaeza:
@luc-demeyer is it OK for you?

ghost:
@luc-demeyer @pedrobaeza This PR is now closed because I didn't followup. The PR https://github.com/OCA/l10n-france/pull/401 in the french localization is also in the same state. The only one still alives is the one from the spanish localization https://github.com/OCA/l10n-spain/pull/2711. 

What should we do ?

pedrobaeza:
You can resubmit another PR with the same branch.

PR #207, opened 2022-12-27
[16.0][MIG] product_harmonized_system

alexis-via:


alexis-via:
It is important to note that, in v16, the delivery module provides a M2O field "country_of_origin" on product.template, cf https://github.com/odoo/odoo/blob/16.0/addons/delivery/models/product_template.py#L14

So it is exactly the same field as our field "origin_country_id" provided by our OCA module "product_harmonized_system" !

Option 1: we adopt the new field "country_of_origin" and add a migration script to move "origin_country_id"  to "country_of_origin"
Option 2: we hide the new field "country_of_origin" in our module "product_harmonized_system_delivery" and we continue to use our good old field. We also set "country_of_origin" as a related field of "origin_country_id", so that they have the same value.

So far, I choose option 2, because of the bad naming of the new field "country_of_origin" (missing "_id" suffix). But, if you guys prefer option 1, it is also ok for me.

alexis-via:
Currently, intrastat_product depends on "sale_stock" and "purchase_stock", but not on "delivery". If we choose option 1, it would require to add a dependency on "delivery". So it's one more argument for option 2.

ajaniszewska-dev (approved):
LG

product_harmonized_system/models/product_product.py
```diff
@@ -0,0 +1,21 @@
...
+# @author Luc de Meyer <info@noviat.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def get_hs_code_recursively(self):
```
ivantodorovich, line 13: This method should be private `_get_hs_code_recursively`
This is a good opportunity to also make private the other similar methods with the same name.

Public model methods are exposed to xmlrpc, it's a good practice to do that only when that's intended.

ivantodorovich (approved):
LG. Non-blocking comment

product_harmonized_system/models/product_product.py
alexis-via: In this case, it is nice to have that kind of method accessible via xmlrpc, isn't it ?

ivantodorovich: But in such case it should return a json instead, or use the decorator `@api.return`.

Otherwise, the response would be something like this:

 ```json
{"result": "hs.code(123,)"}
```

As it is right now, it's better to have it private.

I wuold also argue that, if the idea is to expose this for external services.. it's better to move it to a computed field instead. That would save a roundtrip

sbidoul:
@alexis-via you can retry the merge command now.

Issue #194, opened 2022-10-03
Migration to version 16.0


sbidoul:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0

# Modules to migrate

- [ ] intrastat_base
- [ ] intrastat_product
- [ ] product_harmonized_system

Missing module? Check https://github.com/OCA/maintainer-tools/wiki/%5BFAQ%5D-Missing-modules-in-migration-issue-list

alexis-via:
Decided by @luc-demeyer and me:
In v16, the datas of intrastat.transaction with 2 digits must be moved for localisation modules to the module intrastat_product

alexis-via:
Decided by @luc-demeyer and me:

The 4 fields intrastat_transaction_out_invoice, intrastat_transaction_out_refund, intrastat_transaction_in_invoice, intrastat_transaction_in_refund on res.company are replaced by 2 fields intrastat_transaction_b2b_id and intrastat_transaction_b2c_id.

Field nature of transaction (11, 12, ...) is set on the "code" field of intrastat.transaction. Up until v15, l10n_fr_intrastat_product has "code régime" (11, 21, 25, ...) in the code field on intrastat.transaction and the nature of the transaction (11, 12) in the field "fr_transaction_code" defined in l10n_fr_intrastat_product. In v16, the data must be updated to have the nature of transaction in the "code" field of intrastat.transaction. Do I still need a field for "code régime" ? Maybe it can be deducted from the nature of transaction and the arrival/departure ? Maybe I could set the field "code régime" on the intrastat compute line and not on intrastat.transaction ?

alexis-via:
If it is confirmed that the HS nomenclature is the same between all countries, we should move the HS Code import wizard from l10n_be_intrastat_product to intrastat_product (TODO @luc-demeyer )

alexis-via:
This PR (for v14) is a sort of "proposal" for v16 https://github.com/OCA/intrastat-extrastat/pull/198 I would like it to be the basis for the v16 port.

alexis-via:
In v16, if we move the datas of intrastat.transaction from the localisation modules to intrastat_product, we should remove the company_id field on this object. I really think this object doesn't need a company_id field.

luc-demeyer:
@alexis-via 
based upon my current understanding of the transaction codes in various EU countries I tend to agree, 
[...]

PR #210, opened 2022-12-27
[16.0][MIG] intrastat_base

alexis-via:


alexis-via:
I plan to make an important change here: bool field on fiscal position would be converted to a selection field with B2B/B2C/None as possible values. See https://github.com/OCA/intrastat-extrastat/issues/194 for more.

alexis-via:
done. I'm now working on adapting intrastat_product

alexis-via:
The pre-commit error is unrelated ; it fails in the build of the environment :-(

PR #214, opened 2023-02-08
[16.0][MIG] intrastat_product

alexis-via:
This migration has been made from the v14 PR #198  and the changes of v15 has been re-integrated.
    
As discussed on #194  the intrastat transaction codes are now integrated in the module intrastat_product.  The default value for transaction_id depending on the move_type of the  invoice have been dropped on fiscal position and company: Odoo sets 11 when fiscal position is B2B and 12 when fiscal position is B2C. As a consequence: intrastat.transaction: company_id = False by default.
    
partner_id on computation line is not a related field any more: it allows to auto-fill VAT from partner_id when creating a line manually.


alexis-via:
I have another idea: currently, we have 2 XLS Export buttons : one in the computation tab and a second one in the declaration tab. My idea is to have the XLS Export in the Print menu and/or top menu and use 2 tabs : one for computation lines and another one for the declaration lines. If you think it's not a good idea, please yell NOW !

alexis-via:
Done!
There is now a single XLSX export that has 2 tabs:
1) computation lines
2) declaration lines

There is a single button at the top to generate it.
The declaration type is now mentioned in the XLSX export (surprisingly, nobody complained about it so far!).
Ability to export several declarations in 1 XLSX file: in the tree view, select several declarations and click on "Print > Excel Export".

victoralmau:
Please, cherry-pick https://github.com/OCA/intrastat-extrastat/pull/216 to commit history

alexis-via:
Tests are all green, cool!

intrastat_product/migrations/14.0.1.0.0/pre-migration.py
rven, line 1: Please remove the migrations folder

alexis-via: done. thanks !

intrastat_product/data/intrastat_transaction.xml
```diff
@@ -19,62 +19,88 @@
...
+        <field
+            name="description"
+        >Movements to/from a warehouse (excluding call-off and consignment stock)</field>
     </record>
     <record id="intrastat_transaction_32" model="intrastat.transaction">
         <field name="code">32</field>
-        <field name="description">Supply for sale on approval or after trial (including call-off and consignment stock)</field>
+        <field
+            name="description"
+        >Supply for sale on approval or after trial (including call-off and consignment stock)</field>
```
pedrobaeza, line 42: Why these changes in the migration commit?

alexis-via:
@victoralmau No need to forward port this fix, because my refactor removed and simplified all this code that caused such bug. The method _get_intrastat_country_code() doesn't exist any more here.

intrastat_product/data/intrastat_transaction.xml
alexis-via: In a perfect world...

PR #218, opened 2023-04-06
[12.0][IMP] intrastat_product: Split invoices for memory issues

hildickethan:
- Split invoices every 500 and invalidate cache to help with memory issues on large declarations with large invoices
- Inspired by these changes in 16.0 https://github.com/odoo/odoo/commit/feea4e5f147b3b26cbc032daaf0ceabbe3417912

Personal use case that was raising MemoryError:
- Declaration with ~90000 invoice lines across ~8700 invoices
- Client with many extra fields to manage their invoices
- It would memory error at the base memory limit, with these changes the process length isn't really affected much and no longer memory errors

The diff is quite big but it's all just indenting the `for` inside the split invoices, along with the invalidate_cache at the end and the split invoice `for` itself

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -3,6 +3,7 @@
 # @author Alexis de Lattre <alexis.delattre@akretion.com>
 # @author Luc de Meyer <info@noviat.com>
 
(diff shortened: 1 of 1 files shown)


hildickethan:
closing since it's no longer needed for us and doesn't seem like anyone else needs it

PR #228, opened 2023-09-28
[13.0][FIX]hs_code_link: properly declare OCA hs_code field on product.template

GuillemCForgeFlow:
Before the change, the field was not being updated after the value of the H.S. Code was changed on a Product Template. The related stored field was not working fine with the company-dependent field.
As the `hs_code_id` field is company-dependent, the related `hs_code` value should also be like this and sync the value based on the previous one.

cc @ForgeFlow

### DIFF
--- a/hs_code_link/tests/test_hs_code_link.py
+++ b/hs_code_link/tests/test_hs_code_link.py
@@ -1,0 +1,62 @@
+# Copyright 2023 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo.tests.common import SavepointCase
+
+
+class TestHSCodeLink(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        # Models
+        cls.company_model = cls.env["res.company"]
+        cls.hs_code_model = cls.env["hs.code"]
+        cls.product_template_model = cls.env["product.template"]
+        cls.property_model = cls.env["ir.property"]
+
+        # Instances
+        cls.company_1 = cls._create_company("Company 1")
+        cls.company_2 = cls._create_company("Company 2")
+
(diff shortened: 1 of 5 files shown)


GuillemCForgeFlow:
@leemannd What do you think about this change? 🤔

JordiMForgeFlow (approved):
Code review LGTM

AaronHForgeFlow (approved):
Code review

GuillemCForgeFlow:
@pedrobaeza do you think we can merge this?

pedrobaeza (commented):
Is it the `hs_code` column kept by the ORM in the DB after this change? If so, it will be very confusing. Maybe it's better option to switch to a computed non stored, and thus avoiding all the sync stuff.

GuillemCForgeFlow:
@pedrobaeza The `hs_code` column is not kept as I convert it to a `company_dependent` field and thus the column is removed and the values are kept in the `ir_property` table.

Either way, I agree that making it a computed non-stored field could be a better option. I will make the changes.

GuillemCForgeFlow:
@pedrobaeza Just changed it!

hs_code_link/tests/test_hs_code_link.py
```diff
@@ -0,0 +1,62 @@
...
+
+        cls.product_template = cls._create_product_template("Product Template 1")
+
+    @classmethod
+    def _create_company(cls, name):
+        return cls.company_model.create({"name": name})
+
+    @classmethod
+    def _create_hs_code(cls, local_code="123456789", description=False):
+        return cls.hs_code_model.with_context(skip_check_hs_code_heading=True).create(
```
pedrobaeza: No need of this context

GuillemCForgeFlow: changed, thanks 😄

pedrobaeza (approved):
/ocabot merge major

PR #229, opened 2023-10-02
[13.0][ADD]product_harmonized_system_heading & product_harmonized_system_heading_stock

GuillemCForgeFlow:
New module to have a better structure on H.S. Codes. If installed, H.S. Codes will require a H.S. Heading to be created first.

cc @ForgeFlow

### DIFF
--- a/product_harmonized_system_heading/models/hs_code.py
+++ b/product_harmonized_system_heading/models/hs_code.py
@@ -1,0 +1,63 @@
+# Copyright 2023 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class HSCode(models.Model):
+    _inherit = "hs.code"
+
+    hs_code_heading_id = fields.Many2one(
+        comodel_name="hs.code.heading", string="H.S. Code Heading", readonly=True
+    )
+
+    @api.constrains("local_code")
+    def _check_local_code(self):
+        for hs_code in self:
+            if len(hs_code.local_code) < 4:
+                msg = (
+                    "The H.S. Local Code needs to have at least 4 characters as this "
+                    "indicate the H.S. Heading."
+                )
+                raise ValidationError(_(msg))
+
+    @api.model
+    def _get_hs_code_heading_id(self, vals):
+        """
+        Method used in the `create` and `write` methods of the `hs.code`.
+        Used to get the corresponding H.S. Code Heading based on the `local_code` value
+        of the H.S. Code. Get the existing one, if there is, otherwise create a new one.
+        :return: a modified vals dict with the `hs_code_heading_id` value
+        """
+        HSCodeHeading = self.env["hs.code.heading"]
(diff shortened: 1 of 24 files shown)


GuillemCForgeFlow:
Hi @pedrobaeza 
Whenever you have time, could you tell me what you think about this new module so that we can better structure H.S. Codes?

As seen in www.tariffnumber.com, H.S. Codes will always have a Heading:
image

By having this module, we could have a better structure on those.

intrastat_product/tests/test_brexit.py
```diff
@@ -12,8 +12,10 @@ class TestIntrastatBrexit(IntrastatProductCommon, SavepointCase):
     def setUpClass(cls):
         super().setUpClass()
         cls.inv_obj = cls.env["account.move"]
-        cls.hs_code_whiskey = cls.env["hs.code"].create(
-            {"description": "Whiskey", "local_code": "22083000"}
+        cls.hs_code_whiskey = (
+            cls.env["hs.code"]
+            .with_context(skip_check_hs_code_heading=True)
```
pedrobaeza: This change shouldn't be needed. By default no extra constraint should exist.

pedrobaeza (changes requested):
About the improvements on the existing module, they are OK. About the new module, not sure if very useful, as at the end, you don't manage the HS codes. You just assign them to products. And what is a blocking point is to add extra constraints that requires context tricks in other unrelated modules. The technique if you want to keep such restriction if to not apply it when on test mode (`config["test_enable"]`) and not testing the constraint itself.

GuillemCForgeFlow:
Hi @pedrobaeza 
Thank you for the feedback, I've applied the changes now.

Regarding the use of the module, I've done it in a way that you don't need to install the module to still be using the H.S. Codes provided by the `product_harmonized_system` module. This module only adds a new level of complexity in order to have H.S. Codes in a better structure, just for informational purposes.

We have a customer need that requires us to structure H.S. Codes in such a way and I thought that it could be a good improvement to the current OCA structure. If you don't agree, I will just close it and set it in on a custom module. WDYT?

product_harmonized_system_heading/models/hs_code.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class HSCode(models.Model):
+    _inherit = "hs.code"
+
+    hs_code_heading_id = fields.Many2one(comodel_name="hs.code.heading", readonly=True)
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        if not config["test_enable"] or self.env.context.get("check_hs_code_heading"):
```
pedrobaeza: Instead of this, just do `hs_code_heading_id` a computed writable field, but removing the error if not found.

GuillemCForgeFlow: I've just changed it now to work in the following way:
- When an H.S. Code is created, it will try to find the corresponding Heading, if none is found, it will automatically create a new one, instead of triggering an error to then manually create the Heading.
- If we edit the Local Code, we also call the `_get_hs_code_heading_id` method to get the new Heading.

Doing it like this, we don't need to adapt any tests and by optionally having this module, the user won't need to go through any additional steps.

PR #230, opened 2023-10-05
[16.0] intrastat_product: use shipping partner instead of invoicing partner if any

alexis-via:
After re-reading the spec, I can confirm that, in the case of a triangular operation, we are supposed to use the shipping partner and not the invoicing partner.

Replace @api.onchange by compute methods with store=True, precompute=True and readonly=False for **vat** and **region_code** fields of computation lines.

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -465,15 +465,16 @@ class IntrastatProductDeclaration(models.Model):
             )
         return origin_country
 
-    def _get_vat(self, inv_line, notedict):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why do you need `precompute`?

rlemeur (approved):

alexis-via:
I use precompute=True, readonly=False to make it behave like an onchange.

pedrobaeza:
`precompute=True` is not for that.

alexis-via:
I removed precompute=True, as recommended by Pedro

PR #232, opened 2023-11-07
[15.0][ADD] add new module intrastat_product_hscodes_import

ghost:
This module is based on the installer wizard from l10n_be_intrastat_product in 14.0

The idea of the module is to import all the HS codes defined in data files. Currently the files are from the belgian national bank in french, dutch, german and english. 

The original logic can be found in there https://github.com/OCA/l10n-belgium/blob/14.0/l10n_be_intrastat_product/wizards/intrastat_installer.py

I changed the logic a bit. In the original logic, you had to import the langs one by one with the installer wizard.
In my logic, I check all installed langs and I update all the langs available. 

This PR was discussed with @luc-demeyer and @alexis-via during the OCA Days 2023 and they agreed to move this logic from l10n-belgium to intrastat-extrastat.


ghost:
@alexis-via @luc-demeyer Like discussed, PR ready. All explained in the initial description

luc-demeyer (approved):

PR #234, opened 2023-11-07
[IMP] Update copier template

ghost:


ghost:
@pedrobaeza @luc-demeyer @alexis-via I updated the copier template because my latest merge request (https://github.com/OCA/intrastat-extrastat/pull/232) is failing and it seems to me that updating the pre-commit conf should solve the issue.

pedrobaeza (approved):

PR #235, opened 2023-11-07
[15.0][FIX] intrastat_product: intrastat declaration - restore _compute_numbers that was dropped by …

sergio-teruel:
…PR #149 (removal of intrastat.common

cc @Tecnativa TT43107

ping @carlosdauden @victoralmau

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -33,6 +33,14 @@ class IntrastatProductDeclaration(models.Model):
         )
     ]
 
+    @api.depends("declaration_line_ids")
+    def _compute_numbers(self):
+        for this in self:
+            this.num_decl_lines = len(this.declaration_line_ids)
+            this.total_amount = sum(
+                this.declaration_line_ids.mapped("amount_company_currency")
+            )
+
     @api.model
     def default_get(self, fields_list):
         res = super().default_get(fields_list)
@@ -130,7 +138,7 @@ class IntrastatProductDeclaration(models.Model):
         store=True,
         tracking=True,
     )
-    total_amount = fields.Integer(
+    total_amount = fields.Float(
         compute="_compute_numbers",
         string="Total Fiscal Amount",
         store=True,


intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -196,6 +196,17 @@ def _compute_check_validity(self):
         for this in self:
             this.valid = True
 
+    @api.depends("declaration_line_ids.amount_company_currency")
+    def _compute_numbers(self):
+        for this in self:
+            total_amount = 0  # it is an integer
```
pedrobaeza: Why integer? Isn't that value incorrect and you should remove the fields instead? Are they shown in any place?

xbentan: I think this comment is just to inform the field is wrongly declared in line [133 ](https://github.com/OCA/intrastat-extrastat/pull/235/files#diff-74e1197466e3c9a2c08b006d373e86cca69ac389f34c9e4ed38dc24b84018d80R133) as an integer. Is that also possible to change it as a float and approve the merge? My company need to submit the intrastat report in comming days and we're blocked beacause of this issue : )

pedrobaeza: Correct, but I go further, as if this fields are not shown anywhere, better to simply remove them.

xbentan: both fields are shown in four views:
[form ](https://github.com/OCA/intrastat-extrastat/blob/15.0/intrastat_product/views/intrastat_product_declaration.xml#L75)
[tree](https://github.com/OCA/intrastat-extrastat/blob/15.0/intrastat_product/views/intrastat_product_declaration.xml#L142)
[graph](https://github.com/OCA/intrastat-extrastat/blob/15.0/intrastat_product/views/intrastat_product_declaration.xml#L198)
[pivot](intrastat_product_declaration_view_pivot)

pedrobaeza: OK, then better to convert it to float.

pedrobaeza (approved):
I have adopted here the better fix in #236, but fixing the commit history, message and pre-commit, so let's fix definitively the problem:

/ocabot merge minor

PR #236, opened 2023-11-08
[FIX] total_amount change to Float and [ADD] def _compute_numbers function

xbentan:


pedrobaeza:
As this one contains 2 commits and the pre-commit is invalid, I have cherry-picked the work and continue in #235, so closing.

PR #238, opened 2023-11-09
[16.0][IMP] intrastat_product: Set translate=True to description field from transactions

victoralmau:
Set `translate=True` to description field from transactions

In v16 the transactions in data have been added, it will now be necessary that the localization modules do not add new ones, but it is also convenient to display the corresponding text in each language to avoid duplicate records.

@Tecnativa


luc-demeyer (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #240, opened 2023-11-10
[MIG] intrastat_product_hscodes_import: Migration to 16.0

luc-demeyer:


luc-demeyer:
@victoralmau @pedrobaeza 
This is a module that allows to import the intrastat codes.
We decided during the OCA code sprint to move this code from the belgian localisation to a seperate module.
I use https://www.nbb.be/en/statistics/foreign-trade/nomenclature-and-codes as input.
On this site I have the codes in 4 languages : en, de, fr, nl.
It could be usefull if you add also the Spanish translation.

pedrobaeza:
We already have such import in the Spanish module, but not sure if it's the same source. One question: why putting this in a separate module?

luc-demeyer:
>>>  One question: why putting this in a separate module?

We had a short discussion on this topic during the OCA days.
My initial proposal was to include it in the intrastat_product module.
But many customers seem to prefer a manual encoding of a couple of intrastat codes for their business whereas this wizard imports 10k codes. By making it a seperate module the import option becomes only available to the customer who want this.

Having a seperate module could also make the evolution of the import module easier.
E.g. add webservices to update the common code list, allow to select the codes you want to activate from the list, archive codes who are no longer allowed, add an overview screen of the common code changes compared to the previous year, ...

pedrobaeza:
Although you include this in the main module, the import is optional, so those that don't want to make import are not limited.

About the evolution, it doesn't matter as well: you can do the IMP on this module or the main one. 

I still prefer to have it joined. It's simpler for the migration path from previous versions. Should you accept it if we do a PR merging it?

luc-demeyer:
> Should you accept it if we do a PR merging it?
Ok for me 

@alexis-via 
Do you agree ?

PR #242, opened 2023-11-10
[16.0] intrastat_product: suppl_unit_qty rounded to 1 if between 0 and 0.5

alexis-via:


Doscaal (approved):
Test with data and it works. No more issue and xml_line well generated.

luc-demeyer (approved):

PR #245, opened 2023-11-12
[IMP] intrastat_product - remove default company for intrastat regions

ghost:
Right now, regions have a company by default but if we have two belgian companies or two french companies, it's not logical to have the regions declared in each company nor "user-friendly" to have to edit default regions on the "first" company to have it on each company. 

For example, while migrating l10n_be_intrastat_product and testing on an empty database, my regions were created in one company and not available on the second belgian company. I had to go on the first company to edit the regions and make it available for the second one.

### DIFF
--- a/intrastat_product/models/intrastat_region.py
+++ b/intrastat_product/models/intrastat_region.py
@@ -19,4 +19,4 @@ class IntrastatRegion(models.Model):
     country_id = fields.Many2one(comodel_name="res.country", required=True)
     name = fields.Char(translate=True)
     description = fields.Char()
-    company_id = fields.Many2one("res.company", default=lambda self: self.env.company)
+    company_id = fields.Many2one("res.company")


luc-demeyer (approved):
The localisation modules should set the regions based upon the country_id.
I suggest that the localisation modules inherit from the accounting setup wizard In order to ensure that this is also the case when adding a new company.

pedrobaeza:
There seems to be something weird with the READMEs digest.

ghost:
@pedrobaeza These changes were made by pre-commit. I let it passed since I was not sure it's intended or not. 
Should I remove these ?

pedrobaeza (approved):
I have downloaded the branch locally, undoing the digest changes, and they are not done. I think you have an old tooling in your local. Please update it.

As this is now fixed, merging it:

/ocabot merge patch

PR #246, opened 2023-11-12
[FIX] intrastat_product - set region_code based on code or the region

ghost:
If we don't use the method "_get_region_code" and inherit the method in the localisation module, the region code is never set even when using the default method "_get_region" that fetchs the region based on invoice, warehouse or company.

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -411,7 +411,7 @@ class IntrastatProductDeclaration(models.Model):
         If none found, get the company's default intrastat region.
 
         """
-        region = False
+        region = self.env["intrastat.region"]
         move_type = inv_line.move_id.move_type
         if move_type in ("in_invoice", "in_refund"):
             po_line = self.env["purchase.order.line"].search(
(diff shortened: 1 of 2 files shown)


intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -709,7 +709,7 @@ def _gather_invoices(self, notedict):
                     "amount_accessory_cost_company_currency": 0.0,
                     "transaction_id": intrastat_transaction.id,
                     "product_origin_country_id": product_origin_country.id or False,
-                    "region_code": region_code,
+                    "region_code": region_code or region and region.code or False,
```
pedrobaeza: Please avoid this old-stylish spaghetti code:

```suggestion
                    "region_code": region_code or region.code,
```

pedrobaeza: If `region` may be False, then the method `_get_region` should return an empty recordset.

ghost: @pedrobaeza Code adapted according to your suggestion. Thanks

pedrobaeza (approved):

pedrobaeza:
It seems one test is not satisfied. Please check.

ghost:
@pedrobaeza The issue is fixed

PR #249, opened 2023-11-23
[16.0][FIX] intrastat_product: Do not display missing VAT number error when…

aritzolea:
… fiscal position is B2C

PR #250, opened 2023-11-24
[16.0][FIX] intrastat_product: no VAT warning for B2C

alexis-via:


pedrobaeza:
Isn't this the same as #249 ?

PR #252, opened 2023-11-24
Added encoding check

rosenvladimirov:
Added encoding checking for imported files, at moment use windows-1252, but not all files for import with this encoding.


pedrobaeza:
Incorrect pull request, removing hundred of files.

PR #258, opened 2023-12-29
[17.0][MIG] product_harmonized_system + 2 glue modules

alexis-via:
Migrate **product_harmonized_system** and its 2 glue modules  **product_harmonized_system_stock** and **product_harmonized_system_stock_delivery**

Rename product_harmonized_system_delivery to product_harmonized_system_stock_delivery, because it now depend on the new module stock_delivery.


pedrobaeza (changes requested):
You must separate each module in one pull request, having each commit history for each one.

pedrobaeza:
@alexis-via, you can't merge a PR without reviews, without using ocabot, and not attending changes requested. I'm undoing this right now.

alexis-via:
@pedrobaeza You can't block PRs to oblige contributors to abide by your own rules. There is only 1 main module in this PR. The 2 other modules are just glue module related to the main module. So there is really NO REASON to make 3 PRs.

IF YOU HAVE TIME TO LOOSE TO MAKE 3 SEPARATE PR, PLEASE DO IT !!!

alexis-via:
I don't have time to loose for those stupid rules.

pedrobaeza:
These are not my rules, but OCA rules:

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst

What you can't do is systematically don't follow them on any contribution you perform, and merge even without the ocabot command and not having any review.

pedrobaeza:
> I don't have time to loose for those stupid rules.

Well, with that attitude, I can say the same when I review your PRs and don't even waste any minute of my time.

Contributing is not at any cost. You don't forward-port any PR you do in lower versions (unless they are even versions), your contributions are not OK in form (in content they can be very good, but it's not enough for sharing in a community where following rules assures consistency, knowledge transfer and so on), etc. Please think about this.

alexis-via:
Again, there is only 1 "real" module in this PR. Let's apply those rules intelligently. A rule that multiply by 3 the time spent by contributors to migrate module is a bad rule : either we change it or we try to apply it intelligently.

If we don't apply those rules intelligently, if OCA starts to be an organisation without any intelligence and without good sense, then I'm very pessimistic for the future of OCA and for my role as a contributor in here.

pedrobaeza:
Glue modules are also modules, even if their size is less. Although you don't find utility is splitting the pull request, it's intended due to these reasons:

- Reviews are easier. Right now reviewing 54 files is not the same (even if the changes are minimum) as reviewing only 25.
- It also scares/avoids people to review it, as having non-related things (for example, someone wants to HS system, but don't have delivery stuff) prevents people on getting into.
- Blame operations are a hell. Each time I have to go to do blame and arrives in your infamous refactorization commit touching everything mixed in https://github.com/OCA/bank-payment/commit/bee309384134cde9c4655f89a24c0612ddc84e56 forces me to abandon that investigation way.
- Talking about the commit separation, you are mixing unrelated changes in the `pre-commit auto-fixes` commit (6bd3ffe9009b09114278a8a6136120efdefdf18b), and that's because you are not strictly following the migration guide: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0#technical-method-to-migrate-a-module-from-160-to-170-branch. Not doing this correctly, makes very difficult for your future you or any other people using this module to investigate things through the commit history. Each change must be in their proper place.
- If there's any requested change in any of the glue modules, this will block the merge of the other ones.

About the loss of time, I invite you to do it properly at least 2-3 times, you'll see that is not real, and there's barely more time invested splitting it. You just need to split the efforts into smaller batches.

Another important rule you have broken here is the lack of other reviews. You should embark other colleagues in reviews (real reviews, not only lobby ones), or review other PRs and ask in exchange that they review yours.

And finally, please be a bit generous. I'm still waiting for any of the reviews I have asked from you in the contributions I have made in the modules you are involved (mostly in bank-payment), while I have reviewed plenty of yours.

PR #260, opened 2024-01-23
[16.0] intrastat_product: add confirmed state + several small fixes/improvements

alexis-via:
Add a third step "confirmed" between "draft" and "done". It is necessary to have error messages of XML file generation that refer to declaration lines. It was my idea to generate the declaration lines and the XML at the same time ; but real life experience showed it was not a good idea (sorry for that !). Some error messages that block XML generation could refer to computation lines but not all, because, for the fields that are summed in declaration lines, you really need to check the value in the declaration line and point to it if the value is bad.

Add a new field "line_number" on declaration line, to be used in error message when generating XML file.

Add store=True on 3 computed fields (reporting_level on declaration, src_dest_country_code and product_origin_country_code on computation lines). It fixes a bug on src_dest_country_code and product_origin_country_code on computation lines, when you could not set manually a country code while leaving the m2o field empty.

Improve display_name of intrastat.product.declaration

note field moved from fields.Text to fields.Html

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -46,7 +46,8 @@ class IntrastatProductDeclaration(models.Model):
         comodel_name="res.company",
         string="Company",
         required=True,
-        states={"done": [("readonly", True)]},
+        readonly=True,
+        states={"draft": [("readonly", False)]},
         default=lambda self: self.env.company,
     )
     company_country_code = fields.Char(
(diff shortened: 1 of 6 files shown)


alexis-via:
I also added readonly access to auditor group

hparfr (approved):
code read
LGTM

ljsalvatierra-factorlibre:
Why did you break backwards compatibility removing the method `done`?

https://github.com/OCA/intrastat-extrastat/pull/260/files#diff-74e1197466e3c9a2c08b006d373e86cca69ac389f34c9e4ed38dc24b84018d80L963

PR #264, opened 2024-01-31
[16.0][FIX] intrastat_product: skip invoice with non EU country

MarinaAForgeFlow:
Added check to skip the invoice if the method _get_partner_country has resulted in an error because of the country not being a EU country.
Also added hooks to allow having empty values if the field is not required. For example, in the Danish intrastat report some fields are not required under certain circumstances.

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -636,6 +636,9 @@ class IntrastatProductDeclaration(models.Model):
                 partner_country = self._get_partner_country(
                     inv_line, notedict, eu_countries
                 )
+                if notedict["invoice"][notedict["inv_origin"]]:
+                    continue
+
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):
code LG

MarinaAForgeFlow:
Hi @joao-p-marques. I see that you worked on the intrastat_product migration to v16 and that you refactored the code for the method "_get_partner_country". In this PR I have added a check to skip the invoice if the partner country is outside the EU, however, I am wondering now if my fix is really necessary.  
In the case where the partner country is the same as the company country, the invoice is skipped and the message is logged instead of added to the notedict, as it is not an error but a reason to filter out. Do you believe that this should be done too for the case where the partner country does not belong to the EU?

MarinaAForgeFlow:
Hi @alexis-via  , is there any possibility that we can get this PR re-opened?

PR #265, opened 2024-02-01
[16.0][IMP] intrastat_product: erase computed lines when the declaration scope changes

MarinaAForgeFlow:


pedrobaeza:
This should depend on the mode selected in the declaration.

MarinaAForgeFlow:
Hi @pedrobaeza, could you explain why do you think that it should depend on the mode? What I think is that, if the user changes, for example, the month of the report once the computation lines are already generated, and then confirms and exports, the report the user will get will be incorrect, since the lines will not belong to the period of the report as they have not been erased or recomputed after the change.

pedrobaeza:
Maybe you want to aggregate several months in the same report for your stats...

MarinaAForgeFlow:
Oh I did not think about this use of the report. Then this change may not be needed at all, as the user could want to aggregate arrivals and dispatches too.

PR #268, opened 2024-03-15
[17.0][FIX] product_harmonized_system: Resolve AttributeError in hs_code _compute_display_name

AndreuOForgeFlow:
This PR fixes an error caused by attempting to process non-string types with the `shorten` function. I replaced `shorten` with direct string operations. This change simplifies the code and eliminates the error.

This `shorten` function was added in v15, as seen here: https://github.com/OCA/intrastat-extrastat/commit/5a6f48963cfeb7698379b206c9f08c25577c6539#diff-ce6878dd561fd41665ff42ec76a5b427470893007d3342eb9ab5d42fbd50b9d6R7

Also, I don't think it is necessary to use an external library when the functionality of `shorten` can be covered using string operations.

### DIFF
--- a/product_harmonized_system/models/hs_code.py
+++ b/product_harmonized_system/models/hs_code.py
@@ -84,7 +84,7 @@ class HSCode(models.Model):
     @api.depends("local_code", "description")
     def _compute_display_name(self):
         for this in self:
-            name = this.local_code
+            name = this.local_code or ""
             if this.description:
                 name += " " + this.description
             name = shorten(name, 55)


GuillemCForgeFlow (approved):
LGTM! Makes sense 👍🏿

alexis-via:
Why do you remove the calll to shorten ?
Cutting with [:55] can cut anywhere, in the middle of a word, which is really dirty.
shorten() handle it nicely ; it cuts after the end of a word, and it adds "..." at the end. A nice and readable cut.

alexis-via:
```
>>> from textwrap import shorten
>>> a='Alexis think it should be allowed to port glue modules in the same PR as the main module'
>>> a[:50]
'Alexis think it should be allowed to port glue mod'
>>> shorten(a, 50)
'Alexis think it should be allowed to port [...]'
```

GuillemCForgeFlow:
> Why do you remove the calll to shorten ? Cutting with [:55] can cut anywhere, in the middle of a word, which is really dirty. shorten() handle it nicely ; it cuts after the end of a word, and it adds "..." at the end. A nice and readable cut.

I guess that we can keep the call to the shorten function, I agree that it makes more sense to handle large descriptions in a better way. Nevertheless, what for sure needs to be solved is to initialize the `name` to have an empty string, if not, it fails because the shorten function doesn't handle to return an empty string if it's given a boolean type. (empty `local_code`)

alexis-via:
ok, let's do the fix and keep the call to shorten()

GuillemCForgeFlow:
@alexis-via then, I'd say this is ready to be merged, right? So we can test https://github.com/OCA/intrastat-extrastat/pull/269 with the full flow.
thank you in advance 😄

alexis-via (approved):

PR #261, opened 2024-01-24
[17.0][MIG] intrastat_product

alexis-via:
Migration from v16 PR https://github.com/OCA/intrastat-extrastat/pull/260

There are still 6 ruff errors "B023 Function definition does not bind loop variable `rec`", but I'm not sure if they are relevant or not.

PR #276, opened 2024-05-25
[17.0][MIG] intrastat_product

dreispt:
Supersedes #261


intrastat_product/models/intrastat_transaction.py
```diff
@@ -0,0 +1,37 @@
...
+    code = fields.Char(required=True)
+    description = fields.Text(translate=True)
+    # intrastat.transaction are shared among companies by default
+    company_id = fields.Many2one("res.company")
+    active = fields.Boolean(default=True)
+
+    @api.depends("code", "description")
+    def _compute_display_name(self):
+        for this in self:
+            name = this.code
```
GuillemCForgeFlow:
```suggestion
            name = this.code or ""
```
Following with comment here https://github.com/OCA/intrastat-extrastat/pull/261#discussion_r1540616137. I guess you can add it here.

GuillemCForgeFlow (changes requested):
After this is changed, the rest looks good to be merged IHMO 👍🏿

etobella:
@dreispt I think with a small typo change we can get this merged :wink:

GuillemCForgeFlow:
> @dreispt I think with a small typo change we can get this merged 😉

Well it's not actually a typo, if the change is not added it will fail upon the creation of a new Transaction Type, therefore we need the change first.

image

etobella:
Well, let's say it is small change :wink:

juanmagil:
When will this migration to the 17.0 branch be merged?

intrastat_product/models/intrastat_transaction.py
dreispt: This bust be a pre-existing bug affecting previous versions too.
I added a fix commit.

dreispt:
Lint fixes were added and build is green now.
I believe it can be merged based on the approvals at #261

gurneyalex (approved):
ok for me.

PR #281, opened 2024-09-11
[16.0][ADD] module product_harmonized_system_per_country

emiliesoutiras:
We need to be able to define different HS codes depending on the destination country, both for product categories and for the products themselves.
For example, the HS Code of a given product is different if you ship to the USA or to Canada, even though in both cases the origin is the same.


emiliesoutiras:
@metaminux, @akretion

product_harmonized_system_per_country/models/account_move.py
```diff
@@ -0,0 +1,45 @@
...
+        ):
+            res.setdefault(line.product_id.id, {"weight": 0})
+            vals = self._prepare_intrastat_line_info(line)
+            if vals.get("hs_code_id"):
+                weight = vals.pop("weight")
+                res[line.product_id.id].update(vals)
+                res[line.product_id.id]["weight"] += weight
+            else:
+                res.pop(line.product_id.id)
+        # sort res :
```
Kev-Roche:
```suggestion
```

product_harmonized_system_per_country/report/deliveryslip_report.xml
```diff
@@ -0,0 +1,26 @@
...
+        <field name="inherit_id" ref="delivery.report_delivery_document2" />
+        <field name="priority" eval="1000" />
+        <field name="arch" type="xml">
+            <xpath expr="//t[@t-set='has_hs_code']" position="attributes">
+                <attribute name="t-value">
+                     o.move_ids.filtered(lambda l: l.product_id.hs_code)
+                </attribute>
+            </xpath>
+<!--            <xpath expr="//span[@t-field='']" position="replace">-->
+<!--            </xpath>-->
```
Kev-Roche, line 23:
```suggestion
```

product_harmonized_system_per_country/models/product_product.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def get_hs_code_recursively(self):
+        res = self.env["hs.code"]
+        if self:
+            self.ensure_one()
+            if self.hs_code_id:
+                res = self.hs_code_id.filter_per_country()
+            else:
+                res = super().get_hs_code_recursively()
+        return res
```
Kev-Roche, line 19: It is almost the same method in product.product and product.category, is it possible to do a mixin and avoid duplicate method ?

product_harmonized_system_per_country/models/product_product.py
```diff
@@ -0,0 +1,39 @@
+#  Copyright (c) 2024 Groupe Voltaire
+#  @author Emilie SOUTIRAS  <emilie.soutiras@groupevoltaire.com>
+#  License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
```
Kev-Roche: This class is in product_product.py, It will be clearer in a dedicated file product_template.py (or rename the file as product.py but I prefer the first option)

Kev-Roche (changes requested):

product_harmonized_system_per_country/models/product_product.py
emiliesoutiras: These functions are defined in the product_harmonized_system module, the mixin cannot be done here.

Kev-Roche (approved):

luc-demeyer:
@emiliesoutiras 

Can you provide evidence that this module is required ?

HS Code are defined on a global basis and the 6 digit HS Code is the same everywhere.
Countries can add specific digits. The 'local_code' field has been created to encode these differences.

In the current module, the 'company_id' field on the HS code allows to load different codes per company.
This is indeed something we could improve by making it country_id dependent in stead of  company_id.

I think that such a change may do the job and removes the need for this module.

luc-demeyer (changes requested):
I think we should try to accommodate this requirement into the base 'product_harmonised_system' module (cf.  my comment in the conversation section of this PR).

emiliesoutiras:
> @emiliesoutiras
> 
> Can you provide evidence that this module is required ?
> 
> HS Code are defined on a global basis and the 6 digit HS Code is the same everywhere. Countries can add specific digits. The 'local_code' field has been created to encode these differences.
> 
> In the current module, the 'company_id' field on the HS code allows to load different codes per company. This is indeed something we could improve by making it country_id dependent in stead of company_id.
> 
> I think that such a change may do the job and removes the need for this module.

@luc-demeyer 

Our company currently has products manufactured in France and shipped to the US and Canada, and their HS code is different: for example, for the US, it's 4201.00.1000, and for Canada, it's 4201.00.6000.
We're talking about a single product from a single company, which must have a different HS code depending on the country of destination.

Firstly, the current design of the HS Code doesn't allow multiple HS Codes for the same product/category, so we're proposing a solution with relationships between HS Codes.
Secondly, we are proposing an optional country, so as to have a default HS Code (without company or country), and linked HS Codes that are by country, in order to dynamically find the right HS Code according to the destination country.

Conceptually, I'd ideally imagined a base code containing, for example, 420100 and a local_code per country containing 1000 for the US and 6000 for Canada, but I've adapted to the current design so as not to break everything.

You propose to make the HS Code dependent on the country_id, is there the same mechanism as the company_id dependency in Odoo? Or is it through SQL constraints on the model that you intend to do this?

PR #286, opened 2024-11-06
[17.0][IMP+FIX] intrastat_product: Improve company rules

victoralmau:
Improve company rules

Changes done:
- Improve performance using `[('company_id', 'in', company_ids + [False])]`
- Change `intrastat.product.declaration` rule because `company_id` field is required.
- Add rules for `intrastat.product.declaration.line` and `intrastat.product.computation.line`

@Tecnativa

PR #287, opened 2024-11-07
[17.0][FIX]intrastat_product: fix `migrate` method params

GuillemCForgeFlow:
Fixes migration script introduced in https://github.com/OCA/intrastat-extrastat/pull/286

cc @ForgeFlow

### DIFF
--- a/intrastat_product/migrations/17.0.1.1.0/post-migration.py
+++ b/intrastat_product/migrations/17.0.1.1.0/post-migration.py
@@ -6,5 +6,5 @@ from openupgradelib import openupgrade
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.load_data(
-        env.cr, "intrastat_product", "migrations/17.0.1.1.0/noupdate_changes.xml"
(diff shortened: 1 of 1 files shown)


GuillemCForgeFlow:
@victoralmau @pedrobaeza @carlos-lopez-tecnativa I guess we can have this merged?

pedrobaeza:
Uhm, it was correct before. It has the decorator `openupgrade.migrate()`, and that's why it's `env`, and not `cr`.

GuillemCForgeFlow:
> Uhm, it was correct before. It has the decorator `openupgrade.migrate()`, and that's why it's `env`, and not `cr`.

I see now, strange it is failing for my case. Will check in more detail and get back 👍🏿

GuillemCForgeFlow:
@pedrobaeza I see that starting from v17, the `env` param should be used instead right? https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L297 - https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L320

pedrobaeza (approved):
OK now.

PR #298, opened 2025-02-01
[18.0][MIG] intrastat_product

alexis-via:
What do we do with the new product type "combo" ?
https://github.com/odoo/odoo/blob/18.0/addons/product/models/product_template.py#L64
Show warning ?


alexis-via:
Requires #299

alexis-via:
Tests are green. Ready to merge.

PR #302, opened 2025-03-17
[17.0][FIX] intrastat_product: skip invoice with non EU country

JordiMForgeFlow:
FW of https://github.com/OCA/intrastat-extrastat/pull/264

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -585,6 +585,7 @@ class IntrastatProductDeclaration(models.Model):
     def _gather_invoices_init(self, notedict):
         """placeholder for localization modules"""
 
+    # flake8: noqa: C901
     def _gather_invoices(self, notedict):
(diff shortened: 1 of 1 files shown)


JordiMForgeFlow:
@alexis-via @luc-demeyer could we get a review here? :)

AndreuOForgeFlow (approved):
LGTM!

GuillemCForgeFlow (approved):
lgtm 👌🏿

GuillemCForgeFlow:
@pedrobaeza could we reopen and merge this fix to v17.0? thank you in advance 🙂

alexis-via:
FYI, if you deliver the product to an EU member state but you invoice a customer in a non-EU member state, you are supposed to take the invoice in the intrastat declaration. That's why we don't skip invoice with non EU country ; we look at the fiscal position and the destination country of the invoice. I think the current implementation is OK on this.

I think your change on HS code is good.

PR #303, opened 2025-04-02
[16.0][IMP] intrastat_product: Add menu to avoid redundancy

juancarlosonate-tecnativa:
https://github.com/OCA/l10n-spain/pull/4109
https://github.com/OCA/l10n-france/pull/643

@tecnativa TT55417
@victoralmau @pedrobaeza @alexis-via

### DIFF
--- a/intrastat_product/README.rst
+++ b/intrastat_product/README.rst
@@ -149,6 +149,7 @@ Contributors
 
     * João Marques
     * Víctor Martínez
+    * Juan Carlos Oñate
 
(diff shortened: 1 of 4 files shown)


victoralmau (approved):
It makes sense to be consistent and avoid adding a specific menu for each intrastat localization module.

pedrobaeza (approved):
Yeah, it seems interesting to converge more and more to a common ground.

/ocabot merge patch

PR #305, opened 2025-04-10
[18.0][ADD] product_harmonized_system_tax_rate

hparfr:
set import tax rates per { country, HS Code}


product_harmonized_system_tax_rate/models/product.py
```diff
@@ -0,0 +1,45 @@
...
+        string="H.S. Code Tax Rate",
+        compute="_compute_hs_code_tax_rate_str",
+    )
+
+    @api.depends("origin_country_id", "hs_code_id", "hs_code_id.tax_rate_ids")
+    def _compute_hs_code_tax_rate_id(self):
+        for product in self:
+            product.hs_code_tax_rate_id = product.hs_code_id.tax_rate_ids.filtered(
+                lambda x, prod: x.origin_country_id == prod.origin_country_id
+            )
```
nayatec:
```suggestion
            product.hs_code_tax_rate_id = product.hs_code_id.tax_rate_ids.filtered(
                lambda x: x.origin_country_id == product.origin_country_id
            )[:1]
```

product_harmonized_system_tax_rate/models/hs_code_tax_rate.py
```diff
@@ -0,0 +1,35 @@
...
+
+    hs_code_id = fields.Many2one(
+        "hs.code",
+        string="H.S. Code",
+        required=True,
+        ondelete="restrict",
+        help="Harmonised System Code. Nomenclature is "
+        "available from the World Customs Organisation, see "
+        "http://www.wcoomd.org/. You can leave this field empty "
+        "and configure the H.S. code on the product category.",
```
nayatec:
```suggestion
        help="Harmonised System Code. Nomenclature is "
        "available from the World Customs Organisation, see "
        "http://www.wcoomd.org/.",
```

nayatec: @hparfr The HS Code is mandatory here as we relate it to a country code and a tax rate.
It maybe worth adding a sql constraint to have hscode/country unique. What do you think ?

product_harmonized_system_tax_rate/models/product.py
nayatec: With a unique hscode/country we are fine, but if there is no uniqueness constraint then we may have multiple values. This is why i think having a sql constraint as mentionned above should be better.

product_harmonized_system_tax_rate/models/hs_code_tax_rate.py
```diff
@@ -0,0 +1,35 @@
...
+        string="Country of Origin",
+        required=True,
+        help="Country of origin of the product i.e. product " "'made in ____'.",
+    )
+
+    tax_rate = fields.Float(
+        string="Tax Rate %",
+        required=True,
+        help="Tax rate percentage for this H.S. Code / Country of Origin.",
+    )
```
nayatec, line 41:
```suggestion
    )
    
    _sql_constraints = [
        (
            "hs_code_origin_country_uniq",
            "UNIQUE(hs_code_id, origin_country_id)",
            "H.S. Code/Country of Origin must be unique!",
        ),
    ]
```

product_harmonized_system_tax_rate/models/product.py
hparfr: I think it's subject to issue https://docs.python-guide.org/writing/gotchas/#late-binding-closures

`[:1]` added for mor readability

hparfr:
Thanks for your review @nayatec

bealdav (approved):
Tested, LGTM. Thanks

nayatec (approved):
LGTM, thanks !

PR #306, opened 2025-04-14
[16.0][IMP] intrastat_product_hscodes_import and product_harmonized_system: Update importer, data, and remove company_id 

juancarlosonate-tecnativa:
Discrepancies between the codes data and countries cause various issues when trying to unify multiple declarations, even if it operates through different entities. Both the Combined Nomenclature codes and the 12-digit TARIC codes are standardized across the European Union, with consistent characteristics such as the code itself, measurement units, and description.

Given this, the old importer was relying on unofficial data, which is not appropriate. I have updated the importer to use official data from the TARIC database (https://circabc.europa.eu/ui/group/0e5f18c2-4b2f-42e9-aed4-dfe50ae1263b/library/6e7dc94f-70e4-44ed-9560-b9794979cb87?p=1&n=10&sort=modified_DESC), including the correct hierarchies and attributes.

@tecnativa TT55417
@pedrobaeza @victoralmau @luc-demeyer @alexis-via


victoralmau (commented):
Maybe a migration script should be done to leave all records without a defined company, but I'm not sure, what do you think @pedrobaeza ?

pedrobaeza:
I don't think we should go this way. Instead. on each company, you should make an import of the HS codes. Where you have to be sure that you are looking for the correct HS codes is in report code, not selecting HS codes from other company.

victoralmau:
> I don't think we should go this way. Instead. on each company, you should make an import of the HS codes. Where you have to be sure that you are looking for the correct HS codes is in report code, not selecting HS codes from other company.

Ok, but what happens if you create a product without company (can be used in all) but link a HS code that has a specific company?

pedrobaeza:
That field should be company dependent.

juancarlosonate-tecnativa:
But some of the existing codes are common across countries wouldn't it be redundant to duplicate them?

pedrobaeza:
Yes, but it's the only way to assure everything is correct, and it's how it has been thought in first place.

juancarlosonate-tecnativa:
What would the migration script look like?

pedrobaeza:
This should have migration scripts, and maybe some code to assign to other companies values the corresponding code if exists.

pedrobaeza:
You have this method: https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py

On pre-migration, you should rename the field, and on post-migration, call that method.

juancarlosonate-tecnativa:
Why should the field be renamed if his name hasn't been modified?

pedrobaeza:
Because Odoo removes the column in the update phase, so we don't want to lose data meanwhile. That's what happened at least in previous versions.

juancarlosonate-tecnativa:
@pedrobaeza Doing this causes what we talked about yesterday. In the post-migration im not sure which method you're referring to maybe that's what's failing.

pedrobaeza:
You are doing `rename_fields`, but it should be `rename_columns`, and there's no post-migration script in your code. Another thing to have into account is the duality `product.template`/`product.product`, and you have to check where data is stored. I doubt this is stored in both.

alexis-via (changes requested):
Back in the old days, before v8, hs_code_id on product.template was company_dependant=False.

In v8, the field was set to company_dependent=True, after a suggestion of one of my collegue of Akretion who wanted it for a customer of Akretion. But, after this change, the experience has shown that it was a bad decision based on a bad analysis, so we  decided to switch back to company_dependent=False in this PR dated december 2021:
https://github.com/OCA/intrastat-extrastat/pull/156

An HS code describes the nature of the product ; the nature of the product is the same whatever the company. So the field should stay with company_dependent=False.

But the current implementation has a problem: how to handle properly the local_code, which can be different from one country to another.
Reminder : the 6 first digits of the HS Code is an international nomenclature of the World Customs Organisation (WCO https://www.wcoomd.org/)
Each country can add extra digits after the first 6 digits to fine-tune the international nomenclature to their needs (in the EU, the extension is common to the 27 member stats)
In the OCA datamodel of hs.code :
- hs_code = nomenclature of the WCO, 6 digits only
- local_code = country-specific extension that adds a few digits, often 2.

So we could imagine an improvement of the OCA datamodel on hs.code to be able to set several different country-specific extensions on a single hs.code record.

product_harmonized_system/models/product_category.py
```diff
@@ -14,6 +14,7 @@ class ProductCategory(models.Model):
         "hs.code",
         string="H.S. Code",
         ondelete="restrict",
+        company_dependent=True,
```
luc-demeyer: for a specific product the common code is the same for ALL countries around the world, the last digits can vary per country.
In any case making this company dependent doesn't make any sense.

Imagine you are in a mullti-company setup with e.g. 10 legal entities in spain and shared products and entities in other countries.

Adding a product with his common code will apply to  all of these entities.
The 'local_code' part must be shared between all 10 spanish entities.

This PR should be reworked by changing the local_code part and not via company_dependent but via (to be programmed) 'country_dependent' logic.

juancarlosonate-tecnativa:
What do you think about this, @pedrobaeza , @victoralmau?

This is what i was discussing with Victor making the local HS codes dependent on the company's country code.
I also see a potential issue: we would need to match the local codes with the international nomenclature codes, especially when importing data, like it's done in the l10n_es_intrastat_report module.

luc-demeyer:
Implementing a country based local code must indeed be investigated for impact on add-on modules such as the localisation modules and intrastat_product_hscodes_import, but imho this is the right way to move forward.

juancarlosonate-tecnativa:
@luc-demeyer @alexis-via We're trying to reach a common ground regarding the import process for codes. What operation are you currently using to import the HS codes and local codes?

luc-demeyer:
@juancarlosonate-tecnativa 
cf. module intrastat_product_hscodes_import

juancarlosonate-tecnativa:
@luc-demeyer And where do you get that data from? Do you know if there is a common source at the European level for the national/local codes used by each country?

juancarlosonate-tecnativa:
@luc-demeyer @alexis-via Could you please tell me where you get the data for the codes, including their units of measure, etc.? thx!

luc-demeyer:
@juancarlosonate-tecnativa 
I get the data from the belgian central bank.
They publish the codes in 4 languages (en/fr/nl/de).
I don't think there is a central european database with the local codes for each country.

juancarlosonate-tecnativa:
What do you think about this version? @pedrobaeza @victoralmau @luc-demeyer @alexis-via

pedrobaeza (commented):
Why do we need to have all the XLSX locally instead of downloading them from Internet on demand?

Why not including this directly in the base module instead of using the `import` one?

juancarlosonate-tecnativa:
> Why do we need to have all the XLSX locally instead of downloading them from Internet on demand?
> 
> Why not including this directly in the base module instead of using the `import` one?

I could leave it up to the user to go to the page and download it manually, then add it as a file when selecting the language but this might cause errors.

 I've decided to modify the existing importer to preserve modularity, if the data format changes on the taric files, we'll only need to update the importer without touching the models, i think that's the best approach.

pedrobaeza:
No, what I mean is that the module should download itself the required files, the same as `l10n_es_partner` does:

https://github.com/OCA/l10n-spain/blob/4a61df3ec00aaded2f8c281a92f5a67031713f5f/l10n_es_partner/wizard/l10n_es_partner_wizard.py#L38

About the modularity, as this is a "new feature", I think we can add it in the base module, and keep the previous one as it was, for not breaking current dataset.

juancarlosonate-tecnativa:
ping @pedrobaeza @victoralmau @luc-demeyer @alexis-via

hparfr (changes requested):
Hi, 

>Why not including this directly in the base module instead of using the import one?

Have you considered to put the import feature in a dedicated module instead of the base module ?

Also, adding all these binary files in the git repo seems a bit overkill. (costly on to checkout).

pedrobaeza:
@hparfr there's no other way of getting the data. Please check the ROADMAP and comments. The checkout is not too costly at the end. It was worst with the pandodoc binary files...

About putting in a separate module, at the end, this data is needed, so there's no sense of having 2 modules for the same work.

hparfr:
Where is the roadmap ?
https://github.com/search?q=repo%3AOCA%2Fintrastat-extrastat+roadmap&type=pullrequests

>About putting in a separate module, at the end, this data is needed, so there's no sense of having 2 modules for the same work.

No. In all my business case they never import the full nomenclature. Only the needed codes. It may be 20, it may be 150, or 300. Adding more data is adding noise.
This import feature will be dead code in my case.

pedrobaeza:
Ok, let's include it in the import module.

juancarlosonate-tecnativa:
Code moved to `intrastat_product_hscodes_import`

hparfr:
Thanks @pedrobaeza @juancarlosonate-tecnativa .

I'm ok for the removal of the multi-company rules.

I didn't review the import code.

luc-demeyer:
I think we still have a way to go here:

- harmonised system codes are defined globally and are common to all countries worldwide.
- Europe has added two extra digits which are common to all european countries (Odoo Enterprise adds the table with these 8 digits in their 'account_intrastat' module). 
- Some countries  have added extra digits for national purposes, e.g. Spain, Portugal have added 4 extra digits.

Hence the ideal importer should allow to import the 6 character common code with code description available for all countries around the world. These codes must be shared between all legal entities in a database.

The european (intrastat) importer should add the two extra digits (for which we have created the 'local_code' field).
This local code field should be shared between all european(intrastat) countries (hence country group sharing).

On top of this we have the country specific extensions such as the 4 extra digits for Spain/Portugal.
Imho handling this should be done via the localisation modules and those extensions should be shared between all legal entities belonging to the same country.

The current intrastat/extrastat suite handles only common / local code hence not yet common code / eu extension / country extension with associated global / country group / country  based sharing of the tables.

pedrobaeza:
@luc-demeyer afaik this PR is adding the common European nomenclature, which can be considered the standard one for now. If from other countries want to use it, they can do another importer. We have invested a lot of time, so let's move forward if you don't mind.

luc-demeyer:
@pedrobaeza 
This PR adds the spanisch 12 digit codes.
Most countries use the european 8 digit codes.

The current version adds the european 8 digits and contains translations for 4 languages (eb/fr/de/nl).
We cannot put 12 digit intrastat codes on belgian invoices.

@alexis-via 
what about France ?

juancarlosonate-tecnativa:
> @pedrobaeza This PR adds the spanisch 12 digit codes. Most countries use the european 8 digit codes.
> 
> The current version adds the european 8 digits and contains translations for 4 languages (eb/fr/de/nl). We cannot put 12 digit intrastat codes on belgian invoices.
> 
> @alexis-via what about France ?

You are right, but here we are importing the full codes that are common across Europe.

What differs is how each declaration uses a portion of the code or another, depending on the specific requirements. I believe we should apply that logic individually for each country’s declaration, regardless of importing the complete codes.

Another option would be to compute the code segments, as is currently done with HS codes, and then select the required parts in each country’s declaration according to what is mandatory. https://github.com/OCA/intrastat-extrastat/blob/16.0/product_harmonized_system/models/hs_code.py#L62

pedrobaeza:
@luc-demeyer is the last explanation OK for you?

luc-demeyer:
we could indeed add a _compute_intrastat_code function that returns the EU 8 digit portion but I continue to struggle with this approach (both for spanisch 4 digit extension on top of intrastat and EU 2 digit exension on top of HS Code).

We have around 10000 Intrastat codes in the current version of the importer (data based upon 2023 list published by Belgian authorities).
This is already more than the international 6 digits code list hence when using the hs_code you have multiple to select from.
When we upload the Spanish 12 digits code as the basis to compute the 6 and 8 digits code than we may end up with hundreds of 'duplicates' for the non Spanish part of Europe. 
The second problems comes with the description: every 4 digit extension comes with his own description and I don's see how a _compute_intrastat_description function could come to common description for all of the truncated 8 digits codes.

This being said : a _compute_intrastat_code function is probably to easiest way forward from where we are now. We could combine this with a country-group based sharing of the codes between legal entities and ensure that the web service to retrieve the codes can be easily adapted by the localisation modules so that each country can use a local source to retrieve his codes (hence in Belgium we would retrieve the data from the belgian Natonal Bank and share this between all Belgian Legal entities).

juancarlosonate-tecnativa:
I have revived the idea of ​​one2many, and now the local codes belong to their HS code. They have their own description and country, using the hierarchy of descriptions, etc.

I also adapted the importer so that for countries that only need certain codes, it can be mapped (still needs to be finished).

The downside is that this creates some duplicate codes with different descriptions, and i’m not sure yet how to handle that.

What do you think of this version? @luc-demeyer @victoralmau @alexis-via @pedrobaeza

alexis-via:
After discussing with @juancarlosonate-tecnativa this afternoon at the OCA code sprint in spain, reading the specs of Intrastat for Spain in Spanish (https://sede.agenciatributaria.gob.es/static_files/Sede/Procedimiento_ayuda/DP01/intras03.pdf) and reading various official pages on Taric, etc..., my suggestion to have a solution as simple as possible for spain:
- they write their 12-digit code in the "local_code" field of hs.code
- we add a computed field nc_code on hs.code which would be similar to the hs_code field but get the 8 first digits (instead of the 6 first digits for hs_code)
- I would update the code of l10n_fr_intrastat_product so that, when generating the XML file for intrastat declaration, I will use the new field "nc_code" instead of local_code (on this line of code: https://github.com/OCA/l10n-france/blob/16.0/l10n_fr_intrastat_product/models/intrastat_product_declaration.py#L432
- other country-specific modules would need to be updated too to use nc_code instead of local_code

This solution would comply with the requirements for Spain and keep things as simple as possible.
@luc-demeyer would this solution be ok for you ?

luc-demeyer:
@alexis-via 
I would prefer cn_code (english) in stead of nc_code (french).
I am still a bit puzzled on how we'll handle the description / long_description fields in this 6 vs 8 vs 12 digit complexity. 
But I agree that we should not end up with an over-engineered solution.

alexis-via:
@luc-demeyer you're right, the field name must be in English of course !
I'm glad that you support this simple solution.

alexis-via:
@juancarlosonate-tecnativa do you confirm that you are OK with the proposal ?
If you lack time, I can do a new PR to implement it.

juancarlosonate-tecnativa:
Definitely, the Spanish documentation we discussed earlier, @alexis-via, is not accurate. I was able to confirm this through a call with the Spanish tax agency and by checking the declaration itself: Spain also uses only the first 8 digits of the code. Sorry for the confusion and the wasted time, the documentation is awful and dense, and it can easily lead to misunderstandings.

So, I don't see any need for the one2many since we're all dealing with the same codes. All we're going to do is store the description per language and its hierarchy.

I've also adapted the hierarchy algorithm to be much simpler and avoid duplicates @luc-demeyer and I think with this, everyone should be happy 😄

juancarlosonate-tecnativa:
ping @alexis-via @luc-demeyer Could you please confirm that this looks good to you? Thank you very much!

alexis-via (approved):
I'm ok to remove the company_id field on hs.code

luc-demeyer (changes requested):
In general it looks good.
A few remarks:
- I get a stack trace when trying to load the codes for language nl_NL. We should report a 'missing language' error instead.
- shouldn't we switch from xlrd to openpyxl ? xlsx support has been dropped as from xlrd 2.0.0.

juancarlosonate-tecnativa:
> * I get a stack trace when trying to load the codes for language nl_NL. We should report a 'missing language' error instead.

Done.

> * shouldn't we switch from xlrd to openpyxl ? xlsx support has been dropped as from xlrd 2.0.0.

Openpyxl is not a dependency in Odoo until version 18. Also, xlrd is fixed at version 1.2 in Odoo for now. I don’t see the need for this change at the moment, but I’ll include it in the roadmap.

luc-demeyer (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #307, opened 2025-05-10
[18.0][IMP] product_harmonized_system: add an alternative (long) description to hs.code

nayatec:
It is not always easy to understand exactly which products an HS Code refers to. Adding the possibility for users to include an alternative (long) description, possibly the official explanation of the code, can make their lives easier.

### DIFF
--- a/product_harmonized_system/models/hs_code.py
+++ b/product_harmonized_system/models/hs_code.py
@@ -27,6 +27,13 @@ class HSCode(models.Model):
     description = fields.Char(
         translate=True, help="Short text description of the H.S. category"
     )
+    long_description = fields.Text(
+        string="Alt. Description",
+        translate=True,
+        help="Alternative text description of the H.S. category. "
+        "This can be, for example, the official description of this category "
+        "as defined in the Harmonized System nomenclature.",
+    )
     local_code = fields.Char(
         required=True,
         help="Code used for the national Import/Export declaration. "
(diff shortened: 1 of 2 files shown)


hparfr (approved):

bealdav (approved):

product_harmonized_system/models/hs_code.py
```diff
@@ -27,6 +27,13 @@ class HSCode(models.Model):
     description = fields.Char(
         translate=True, help="Short text description of the H.S. category"
     )
```
luc-demeyer, line 29: H.S. category -> H.S. Code

product_harmonized_system/models/hs_code.py
```diff
@@ -27,6 +27,13 @@ class HSCode(models.Model):
...
+    long_description = fields.Text(
+        string="Alt. Description",
+        translate=True,
+        help="Alternative text description of the H.S. category. "
```
luc-demeyer: H.S. category -> H.S. Code

product_harmonized_system/models/hs_code.py
```diff
@@ -27,6 +27,13 @@ class HSCode(models.Model):
...
+        "This can be, for example, the official description of this category "
```
luc-demeyer: category -> code

luc-demeyer (changes requested):
Since the beginning the help text of the description field is talking about category whereas it should be code.
I suggest we use this PR to fix this historical mistake.

luc-demeyer (approved):

alexis-via (approved):
For the string of the field long_description, I would prefer "Detailed Description" instead of "Alt. Description"

nayatec:
@alexis-via String (and help) changed to "Detailed" instead of "Alt."

PR #308, opened 2025-06-16
16.0 add product harmonized system per country

emiliesoutiras:
Replace PR https://github.com/OCA/intrastat-extrastat/pull/281
We need to be able to define different HS codes depending on the destination country, both for product categories and for the products themselves.
For example, the HS Code of a given product is different if you ship to the USA or to Canada, even though in both cases the origin is the same.


emiliesoutiras:
@meta-it and @akretion

emiliesoutiras:
@metaminux, @kev-roche

metaminux (approved):

PR #309, opened 2025-07-03
[17.0][FIX] intrastat_product: Use Odoo's VAT validation instead of stdnum.vatin.is_valid

AndreuOForgeFlow:
This replaces the use of `is_valid` from the `stdnum.vatin` library with Odoo's validation logic in the VAT constraint check.

The `stdnum.vatin.is_valid()` function expects VAT numbers to be in international format, i.e., prefixed with the country code (e.g., `ESB66676008`), and will return False if the VAT is not prefixed. In contrast, the validation that's used in Odoo (https://github.com/odoo/odoo/blob/211629b232861b960f6650f980d1e3b6acac94d9/addons/base_vat/models/res_partner.py#L184) supports local VAT numbers (e.g., `B66676008`) without requiring the country prefix, because it combines the VAT value with the partner's `country_id`.

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -8,7 +8,6 @@ from collections import defaultdict
 from datetime import date
 
 from dateutil.relativedelta import relativedelta
-from stdnum.vatin import is_valid
 
 from odoo import Command, _, api, fields, models
 from odoo.exceptions import RedirectWarning, UserError, ValidationError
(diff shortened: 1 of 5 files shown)


intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -1129,8 +1128,17 @@ def _compute_product_origin_country_code(self):
     @api.constrains("vat")
     def _check_vat(self):
         for this in self:
-            if this.vat and not is_valid(this.vat):
-                raise ValidationError(_("The VAT number '%s' is invalid.") % this.vat)
+            if not this.vat:
+                continue
+            country = this.partner_id.commercial_partner_id.country_id
```
pedrobaeza, line 1134: The country can be deduced from the initial letters of the vat too. Please follow whole Odoo core code.

AndreuOForgeFlow: Yes, that's the point of this change. The other validation, that uses the lib `stdnum.vatin.is_valid`, only took into account VATs with the prefix (e.g., `ESB66676008`), which is the international format. The validation used in Odoo core code supports local VAT numbers because it (e.g., `B66676008`) because it combines the VAT value with the partner's `country_id`.

pedrobaeza (approved):

GuillemCForgeFlow (approved):
makes sense 👍🏿

pedrobaeza:
Please fw-port it to 18:

/ocabot merge patch

PR #315, opened 2025-09-11
[UPD] intrastat_delivery: Added a different priority order to set the Incoterm: first from the partner, then from the delivery method, and finally from the general settings; if none are set, it will remain empty.

dferest:
Added a different priority order to set the Incoterm: first from the partner, then from the delivery method, and finally from the general settings; if none are set, it will remain empty.

### DIFF
--- a/intrastat_delivery/models/sale_order.py
+++ b/intrastat_delivery/models/sale_order.py
@@ -1,17 +1,59 @@
 # © 2023 FactorLibre - Aritz Olea <aritz.olea@factorlibre.com>
-from odoo import models
+from odoo import api, fields, models
 
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    incoterm = fields.Many2one(
+        "account.incoterms",
+        compute="_compute_incoterm",
+        store=True,
+        readonly=False,
+    )
+
+    def _get_sale_type_incoterm(self):
+        """Check if type_id (sale_order_type module) exists and
+        consider its incoterm  if is defined."""
+        self.ensure_one()
+        if "type_id" in self._fields and self.type_id:
+            return getattr(self.type_id, "incoterm_id", False)
+        return False
+
+    def _get_default_incoterm(self):
+        self.ensure_one()
+        sale_type_inc = self._get_sale_type_incoterm()
+        return (
+            sale_type_inc
+            or self.partner_id.sale_incoterm_id
+            or self.carrier_id.incoterm
+            or self.company_id.incoterm_id
+            or self.env["account.incoterms"]
+        )
+
+    @api.depends(
+        "partner_id",
+        "partner_id.sale_incoterm_id",
+        "carrier_id",
+        "carrier_id.incoterm",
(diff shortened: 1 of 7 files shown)


dferest:
The intrastat_delivery module establishes the Incoterm following this priority:

- First, the Incoterm defined on the partner.
- Second, the Incoterm set on the delivery method (carrier).
- Third, the company configuration Incoterm.
- If none are defined, the field remains empty.

intrastat_delivery/models/sale_order.py
```diff
@@ -1,27 +1,59 @@
...
+
+    def _get_default_incoterm(self):
+        self.ensure_one()
+        sale_type_inc = self._get_sale_type_incoterm()
+        return (
+            sale_type_inc
+            or self.partner_id.sale_incoterm_id
+            or self.carrier_id.incoterm
+            or self.company_id.incoterm_id
+            or self.env["account.incoterms"].browse()
```
aritzolea: .browse() could be removed, `self.env["account.incoterms"]` is enough

intrastat_delivery/models/sale_order.py
```diff
@@ -1,27 +1,59 @@
...
+    )
+    def _compute_incoterm(self):
+        for order in self:
+            order.incoterm = order._get_default_incoterm()
+
     def _action_confirm(self):
-        ret = super()._action_confirm()
+        res = super()._action_confirm()
         for order in self:
+            inc = order._get_default_incoterm()
```
aritzolea: inco/incoterm better?

intrastat_delivery/models/sale_order.py
```diff
@@ -1,27 +1,59 @@
...
+            intr = (
```
aritzolea: intr_trans/transport better?

intrastat_delivery/models/sale_order.py
```diff
@@ -1,27 +1,59 @@
...
-
-    def _prepare_invoice(self):
-        ret = super()._prepare_invoice()
-        ret.update(
-            {
-                "invoice_incoterm_id": self.incoterm.id,
-                "intrastat_transport_id": self.intrastat_transport_id.id,
-            }
-        )
-        return ret
```
aritzolea: Why delete this?

intrastat_delivery/models/sale_order.py
dferest: Solved.

intrastat_delivery/models/sale_order.py
dferest: Improve applied

intrastat_delivery/models/sale_order.py
dferest: Improve applied

intrastat_delivery/models/sale_order.py
dferest: My fault. Solved.

aritzolea (approved):

PR #321, opened 2025-10-30
[16.0][FIX] intrastat_product: set the correct country code for Greece

almumu:
Check correct country code for Greece. The correct is EL instead of GR.

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -17,6 +17,17 @@ from odoo.tools import float_is_zero
 _logger = logging.getLogger(__name__)
 
 
+SRC_DEST_COUNTRY_CODE_MAPPING = {
+    "GB": "XI",
+    "GR": "EL",
+}
+
+PRODUCT_ORIGIN_COUNTRY_CODE_MAPPING = {
+    "GB": "XU",
+    "GR": "EL",
+}
+
+
 class IntrastatProductDeclaration(models.Model):
     _name = "intrastat.product.declaration"
     _description = "Intrastat Product Declaration"
@@ -1167,8 +1178,7 @@ class IntrastatProductComputationLine(models.Model):
     def _compute_src_dest_country_code(self):
         for this in self:
             code = this.src_dest_country_id and this.src_dest_country_id.code or False
-            if code == "GB":
-                code = "XI"  # Northern Ireland
+            code = SRC_DEST_COUNTRY_CODE_MAPPING.get(code, code)
             this.src_dest_country_code = code
 
     @api.depends("product_origin_country_id")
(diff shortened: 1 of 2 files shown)


intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -1169,6 +1169,7 @@ def _compute_src_dest_country_code(self):
             code = this.src_dest_country_id and this.src_dest_country_id.code or False
             if code == "GB":
```
pedrobaeza: Shouldn't you extract this into the new method?

almumu: yes, but then i  have to create two methods, one for src_dest_country_code and one for product_origin_country code,  because the code for GB is not the same in _compute_src_dest_country_code (XI) as the one in _compute_product_origin_country_code (XU)

pedrobaeza: Oh, I see. I think it's legit to create both methods, don't you think? Other opinions are welcome @victoralmau

intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -1267,6 +1269,12 @@ def _prepare_declaration_line(self):
         vals["amount_company_currency"] = int(round(vals["amount_company_currency"]))
         return vals
 
+    def _check_country_code(self, code):
+        """Check and adapt country code for Intrastat purpose"""
+        if code == "GR":
+            code = "EL"
+        return code
+
```
victoralmau:
```suggestion
    _country_code_mapped = {
    "GB": "XU",
    "GR": "EL",
    }


    def _check_country_code(self, f_name):
        # Common method for calling compute_product_origin_country_code() and compute_src_dest_country_code()
        self.ensure_one()
        code = self[f_name] and self[f_name].code or False
        if code and code in self._country_code_mapped():
            code = self._country_code_mapped[code]
        return code

```

victoralmau (changes requested):

intrastat_product/models/intrastat_product_declaration.py
almumu: @victoralmau  @pedrobaeza  maybe we can manage this without a method, and preserve the XU and XI codes, can you take a look?

aritzolea (approved):

victoralmau (approved):
Although perhaps only one mapping could be defined, the change is consistent.

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
Please fw-port it to upper versions.

pedrobaeza:
@almumu are you going to fw-port the changes?

almumu:
@pedrobaeza fw-port done to v17, the tests were not passing with a simple cherry pick,so i have adapted it for v17. Maybe when https://github.com/OCA/intrastat-extrastat/pull/324 is merged i can do a direct fwport with a cherry-pick from v17 to v18.

PR #333, opened 2026-03-09
[17.0][IMP] intrastat_base: Define an appropriate "default" value for fiscal positions in account.chart.template

victoralmau:
Define an appropriate "default" value for fiscal positions in account.chart.template

It is important to define the appropriate value when defining an accounting plan for a company or, for example, when using `account_chart_update` module

@Tecnativa TT61423

### DIFF
--- a/intrastat_base/models/__init__.py
+++ b/intrastat_base/models/__init__.py
@@ -1,4 +1,5 @@
 from . import product_template
 from . import res_company
+from . import account_chart_template
 from . import account_fiscal_position
 from . import account_move
--- a/intrastat_base/models/account_chart_template.py
+++ b/intrastat_base/models/account_chart_template.py
@@ -1,0 +1,21 @@
+# Copyright 2026 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import models
+
+from odoo.addons.account.models.chart_template import template
+
+
+class AccountChartTemplate(models.AbstractModel):
+    _inherit = "account.chart.template"
+
+    @template(model="account.fiscal.position")
+    def _get_account_fiscal_position(self, template_code):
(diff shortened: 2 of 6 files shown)


pedrobaeza (commented):
Why not putting the default here?

https://github.com/Tecnativa/intrastat-extrastat/blob/f439a4c3d9c63aa3d50645f5cad49868dfb5c56c/intrastat_base/models/account_fiscal_position.py#L12

victoralmau:
> Why not putting the default here?
> 
> https://github.com/Tecnativa/intrastat-extrastat/blob/f439a4c3d9c63aa3d50645f5cad49868dfb5c56c/intrastat_base/models/account_fiscal_position.py#L12

I have added the default to the field, but that is not sufficient in the context of the values that will be obtained using `account.chart.template`, because the fiscal positions defined by each chart_template will not have any value defined for the intrastat field, which will be problematic if we want to “update” (restore the values of the fiscal positions) the data.

pedrobaeza:
But what you are talking is about existing ones, not new ones. For existing ones, I would do a migration script doing `UPDATE account_fiscal_position SET intrastat='no' WHERE intrastat IS NULL`, and that's it.

victoralmau:
> But what you are talking is about existing ones, not new ones. For existing ones, I would do a migration script doing `UPDATE account_fiscal_position SET intrastat='no' WHERE intrastat IS NULL`, and that's it.

I will explain a more realistic example without SQL; let's assume that the fiscal positions are defined correctly and for whatever reason (error or ignorance) someone changes some data in the Intrastat field of the fiscal positions. If we use `account_chart_update`, we will expect it to update to the “correct values,” which is why the `account.chart.template` model is necessary.

pedrobaeza (approved):
Understood, thanks.

carlos-lopez-tecnativa (approved):

PR #334, opened 2026-03-13
[17.0][FIX] intrastat_product: Convert fields to Float so as not to lose decimal places

victoralmau:
Convert fields to Float so as not to lose decimal places

Related to https://github.com/OCA/l10n-spain/pull/4901

@Tecnativa TT61473


pedrobaeza (approved):
The original message about this in e7be812a35e35329fa9e30e87d5c0e5fd9b6a002 is:

```
total_amount is a sum of integers, so it should be an integer
```

but that's not true (or not anymore), so it's OK to change it.

christian-ramos-tecnativa (approved):
LGTM

PR #335, opened 2026-03-18
[18.0][FIX] intrastat_product: Convert fields to Float so as not to lose decimal places

victoralmau:
FWP from 17.0: https://github.com/OCA/intrastat-extrastat/pull/334

Convert fields to Float so as not to lose decimal places

Related to https://github.com/OCA/l10n-spain/pull/4901

Please @pedrobaeza and @christian-ramos-tecnativa can you review it?

@Tecnativa TT61473

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -133,9 +133,10 @@ class IntrastatProductDeclaration(models.Model):
         store=True,
         tracking=True,
     )
-    total_amount = fields.Integer(
+    total_amount = fields.Float(
         compute="_compute_numbers",
         string="Total Fiscal Amount",
+        digits="Account",
         store=True,
         help="Total fiscal amount in company currency of the declaration.",
     )
@@ -1314,12 +1315,13 @@ class IntrastatProductDeclarationLine(models.Model):
         string="Suppl. Unit",
         help="Intrastat Supplementary Unit",
     )
-    weight = fields.Integer(help="Net weight in Kg")
+    weight = fields.Float(digits="Stock Weight", help="Net weight in Kg")
     suppl_unit_qty = fields.Integer(
         string="Suppl. Unit Qty", help="Supplementary Units Quantity"
     )
-    amount_company_currency = fields.Integer(
+    amount_company_currency = fields.Float(
         string="Fiscal Value",
+        digits="Account",
         help="Amount in company currency to write in the declaration. "
         "Amount in company currency = amount in invoice currency "
         "converted to company currency with the rate of the invoice date.",


pedrobaeza (approved):

christian-ramos-tecnativa (approved):

alexis-via:
In France, the amount for intrastat declaration are rounded, that's why amounts in declaration line are rounded. I supposed it is the same in BE and NL, otherwise @luc-demeyer would have insisted to have float instead of integer when we initially designed it.
Could you point us to the exact specs in spanish ?

alexis-via:
Specs for France are here : https://www.douane.gouv.fr/sites/default/files/2025-12/22/Note%20de%20r%C3%A9f%C3%A9rence%20pour%20la%20r%C3%A9ponse%20%C3%A0%20l%E2%80%99enqu%C3%AAte%20statistique%20mensuelle%20sur%20les%20%C3%A9changes%20de%20biens%20intra-Union%20europ%C3%A9enne%20%28EMEBI%29%20%202026%20.pdf
page 21 section 3.a "Elle doit être indiquée en euros pour chaque ligne de déclaration en arrondissant à l’unité entière la plus proche." -> it means amounts should be rounded to the nearest integer.

As intrastat is based on EU legislation, I would be very surprised if it would be different for spain than for other EU countries.

pedrobaeza (changes requested):
On hold till confirmation, and if different, the solution may be to keep it as float, but round to integer parametrically depending on the country.

alexis-via:
yes, it could be a solution. But I would be really curious to see the specs of the Spanish intrastat where is says that amounts to declare are with cents. Intrastat specs are supposed to be the same accross all EU, even if there are small differences among countries (maybe it is one of them).
We've been using this datamodel with integer for amounts to declare since v8, so I suppose that many EU countries use it without problems. But of course, I'm open to change and adaptations. I just want to take the right decision without rushing.

pedrobaeza:
Yes, that's why I put it on hold while we check.

pedrobaeza:
Anyway, the other fields that this is aggregating are already float.

alexis-via:
It is normal that fields on intrastat.product.computation.line are floats for amount, weight and supplementary units.

Then, the values of the computation lines are aggregated and then rounded to the nearest integer to generate intrastat.product.declaration.line. We do that because the official specifications (at least for FR and all the other localization that @luc-demeyer maintains for intrastat, at least BE and NL) require the amounts to be in euros without cents and required weight in kg without decimals and the same for supplementary units.

Could you send us a pointer to the official specification for Spain ? As I can read spanish, you give just give the link to the official government website where the specs are. I would be extremely surprised if spain wants the amount, weight and supplementary units with decimals in the intrastat declaration.

pedrobaeza:
@victoralmau can you please locate the specs?

luc-demeyer:
Belgian XML format, cf. https://www.nbb.be/doc/dd/onegate/data/docxml_en.pdf

amount field: 
In the above Belgian manual the examples show the invoice value in integers (cf. field EXTXVAL) hence using integers is safe and ok.
But the XSD allows float and the documentation states:
EXTXVAL invoice value EURO-denominated value

weight field: 
in Belgian doc again no specs on float vs integer but here the sample XML shows two digits.

Odoo Enterprise, module l10n_be_intrastat : 
    <Dim prop="EXTXVAL" t-out="'%.2f' % item['value']"/>
    <Dim prop="EXWEIGHT" t-out="'%.2f' % item['weight']"/>

Hence PR is ok for Belgium, we could be a bit more safe with hardcoded 2 digits for weigth in stead of digits="Stock Weight", but based upon the specs I think that the Belgian declaration with more weight digits will also be ok.

@alexis-via 
Did you try if the French authorities accepts amounts and weight in float format ?
If not than I think that the l10n_fr_intrastat_product module is the place to perform the rounding to integers.

I have no information on other countries for these fields, hence my recommendation to merge this PR is based only on Belgium.

luc-demeyer (approved):

victoralmau:
Spanish documentation https://sede.agenciatributaria.gob.es/static_files/Sede/Procedimiento_ayuda/DP01/INT_FICHERO.pdf (page 4). Values with decimals.

pedrobaeza (approved):
Then it's clear that France seems the exception, not the rule, so this should be merged and `l10n_fr_intrastat` adapted to round the values. @alexis-via can you change your review status?

pedrobaeza:
Oh, I see that there's no "requested changes" from Alexis, so let's merge it then.

/ocabot merge minor

alexis-via:
I'm ok for the change.

But I strongly disagree with the immediate merge, which breaks l10n_fr_intrastat_product. You could have waited a few days for me to prepare a PR for OCA/l10n-france to adapt l10n_fr_intrastat_product, and then we would have been able to merge the PR here and the PR in localization modules at the same time. Now, I have to make the PR for l10n_fr_intrastat_product in emergency :-(

alexis-via:
I'm now looking at the code in more details. Instead of using Fields.Float() for amount fields, we could have used fields.Monetary associated with company currency. It would have been a nicer display in the user interface. Could you make a PR to improve this ?
On my side, I'll do the PR on l10n_fr_intrastat_product now.

intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -133,9 +133,10 @@ def default_get(self, fields_list):
         store=True,
         tracking=True,
     )
-    total_amount = fields.Integer(
+    total_amount = fields.Float(
         compute="_compute_numbers",
         string="Total Fiscal Amount",
+        digits="Account",
```
alexis-via, line 139: for your information, digits="Account" doesn't exist any more. Anyway, we should use fields.Monetary associated with company currency

alexis-via:
You forgot to remove the code that rounds the amount and weight !
Cf https://github.com/OCA/intrastat-extrastat/blob/18.0/intrastat_product/models/intrastat_product_declaration.py#L1264

You must remove the following code in _prepare_declaration_line():
```
        for field in fields_to_sum:
            vals[field] = int(round(vals[field]))
        # the intrastat specs say that, if the value is between 0 and 0.5,
        # it should be rounded to 1
        if not vals["weight"]:
            vals["weight"] = 1
        if vals["intrastat_unit_id"] and not vals["suppl_unit_qty"]:
            vals["suppl_unit_qty"] = 1
        vals["amount_company_currency"] = int(round(vals["amount_company_currency"]))
```

Seems to me that this PR needs a second iteration.

pedrobaeza:
Thanks for the heads up, Alexis. @victoralmau can you check?

alexis-via:
I am surprised that you kept suppl_unit_qty as fields.Integer() on intrastat.product.declaration.line, https://github.com/OCA/intrastat-extrastat/blob/18.0/intrastat_product/models/intrastat_product_declaration.py#L1319

It is on purpose or you forgot to switch this field to float ?

alexis-via:
It would be also an opportunity to switch all the fields that still use digits="Account" to fields.Monetary:

-  'total_amount' on intrastat.product.declaration
- 'amount_company_currency' and 'amount_accessory_cost_company_currency' on intrastat.product.computation.line

PR #342, opened 2026-03-27
[18.0][IMP] intrastat_base: Define an appropriate "default" value for fiscal positions in account.chart.template

victoralmau:
FWP from 17.0: https://github.com/OCA/intrastat-extrastat/pull/333

Define an appropriate "default" value for fiscal positions in account.chart.template

It is important to define the appropriate value when defining an accounting plan for a company or, for example, when using `account_chart_update` module

@Tecnativa TT61423

### DIFF
--- a/intrastat_base/models/__init__.py
+++ b/intrastat_base/models/__init__.py
@@ -1,4 +1,5 @@
 from . import product_template
 from . import res_company
+from . import account_chart_template
 from . import account_fiscal_position
 from . import account_move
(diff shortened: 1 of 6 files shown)


pedrobaeza (approved):

pedrobaeza:
```
2026-03-27 07:41:39,954 306 WARNING odoo odoo.addons.account.models.chart_template: Incorrect usage of try_loading without a fully loaded registry. This could lead to issues. (My Company (San Francisco)-test) 
```

Switch to something like https://github.com/OCA/l10n-spain/blob/bcf130338225628c0326c9b999ebde0d58750313/l10n_es_aeat/tests/test_l10n_es_aeat_mod_base.py#L51

carlos-lopez-tecnativa (approved):

PR #346, opened 2026-04-29
[FIX] intrastat_delivery: Losing incoterm when order does not have a …

dalonsofl:
### Module

intrastat_delivery

### Describe the bug

If a sale order without a carrier has an incoterm and you try to confirm it, after that, the order loses the incoterm because it tries to add the carrier one during the confirmation.

### To Reproduce

1. Create an order with different products and invoice policy based on "Delivered quantities".

(image from runboat OCA)

image

Confirming the order

image

With this change if an order has not a carrier, the incoterm will remain the same .

**Affected versions:**

16.0


AlhMaria (approved):
Hi @dalonsofl 
Looks Good To Me!

PR #347, opened 2026-05-08
[17.0][IMP] intrastat_product: Add a message if there is no weight

victoralmau:
Add a message if there is no weight

Similar to what existed in v15 https://github.com/OCA/intrastat-extrastat/blob/a3ace05118668c3d2caaf257ee6cdca49a3b656b/intrastat_product/models/intrastat_product_declaration.py#L404 and removed in https://github.com/OCA/intrastat-extrastat/commit/3ec9eddbf310ba3be247f72f7ec1532d2b8be404#diff-74e1197466e3c9a2c08b006d373e86cca69ac389f34c9e4ed38dc24b84018d80L414

@Tecnativa TT62349

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -381,6 +381,9 @@ class IntrastatProductDeclaration(models.Model):
             weight = product_weight * source_uom._compute_quantity(
                 line_qty, product.uom_id
             )
+            if not weight:
+                msg = _("Missing weight on product %s.") % product.display_name
+                notedict["invoice"][notedict["inv_origin"]].add(msg)
         else:
             msg = _(
                 "Conversion from unit of measure <em>%(source_uom)s</em> to "


pedrobaeza:
Is the weight mandatory?

victoralmau:
> Is the weight mandatory?

IMO the issue isn't whether it's mandatory or not (I'm not sure if it is, and besides, the requirements may vary by country); the point here is clear: display the message whenever the weight is 0 (just as it was in v15).

pedrobaeza:
Yeah, but this can be very verbose if it's not mandatory and a company is not putting any weight, not showing "real" errors, so better to adjust it according if mandatory or not.

victoralmau:
> Yeah, but this can be very verbose if it's not mandatory and a company is not putting any weight, not showing "real" errors, so better to adjust it according if mandatory or not.

In my opinion, changing the way the message is displayed and whether or not it is made “mandatory” doesn't really matter, because following that same logic, we would also have to change the rest of the messages currently displayed if we consider them to be “mandatory”; in any case, it is merely **INFORMATIVE**, and nothing prevents users from manually adjusting the weight of the lines in the declaration.

pedrobaeza (approved):
Well, I think that needs to be re-thought, but not blocking now.

carlos-lopez-tecnativa (approved):
LGTM

PR #352, opened 2026-05-14
[18.0][IMP] intrastat_product: Add a message if there is no weight

victoralmau:
FWP from 17.0: https://github.com/OCA/intrastat-extrastat/pull/347

Add a message if there is no weight

Similar to what existed in v15 https://github.com/OCA/intrastat-extrastat/blob/a3ace05118668c3d2caaf257ee6cdca49a3b656b/intrastat_product/models/intrastat_product_declaration.py#L404 and removed in https://github.com/OCA/intrastat-extrastat/commit/3ec9eddbf310ba3be247f72f7ec1532d2b8be404#diff-74e1197466e3c9a2c08b006d373e86cca69ac389f34c9e4ed38dc24b84018d80L414

@Tecnativa TT62349

### DIFF
--- a/intrastat_product/models/intrastat_product_declaration.py
+++ b/intrastat_product/models/intrastat_product_declaration.py
@@ -419,6 +419,9 @@ class IntrastatProductDeclaration(models.Model):
             weight = product_weight * source_uom._compute_quantity(
                 line_qty, product.uom_id
             )
+            if not weight:
+                msg = self.env._("Missing weight on product %s.", product.display_name)
+                notedict["invoice"][notedict["inv_origin"]].add(msg)
         else:
             msg = _(
                 "Conversion from unit of measure <em>%(source_uom)s</em> to "


pedrobaeza (approved):
CI error is unrelated and it's an OCA CI infrastructure problem.

intrastat_product/models/intrastat_product_declaration.py
```diff
@@ -419,6 +419,9 @@ def _get_weight_and_supplunits(self, inv_line, hs_code, notedict):
             weight = product_weight * source_uom._compute_quantity(
                 line_qty, product.uom_id
             )
+            if not weight:
+                msg = _("Missing weight on product %s.") % product.display_name
```
carlos-lopez-tecnativa: Non-blocking comment: in v18, you could use
```suggestion
                msg = self.env._("Missing weight on product %s.", product.display_name)
```

carlos-lopez-tecnativa (approved):

PR #353, opened 2026-06-10
[16.0][FIX] intrastat_product: don't abort computation on invalid partner VAT

aritzolea:


almumu (approved):

luc-demeyer:
@aritzolea 
I am trying to understand the use case behind this PR.
If you create a partner record with an invalid VAT number you get:

The VAT number [123456] for partner [Boulangerie] does not seem to be valid. 
Note: the expected format is FR23334175221

Hence the creation of incorrect VAT numbers is already prohibited.

PR #311, opened 2025-08-05
[18.0][MIG] intrastat_product_hscodes_import

JordiMForgeFlow:
res.config.installer was removed https://github.com/odoo/odoo/pull/162435, now using a simple wizard available from intrastat configuration

Including FW of https://github.com/OCA/intrastat-extrastat/pull/306 + transition to use openpyxl instead of xldr (newer version does not support xlsx anymore)

PR #357, opened 2026-09-09
[18.0][FIX] product_harmonized_system: Restore the local code uniqueness

carlosdauden:
Removing company_id from hs.code (https://github.com/OCA/intrastat-extrastat/pull/311) also dropped the whole _sql_constraints block, and with it local_code_company_uniq, without any replacement. Since then nothing prevents two H.S. codes from sharing the same local code, which silently breaks the declarations grouping their lines by that code. The constraint comes back as local_code_uniq, now without company_id, as the codes are standard at EU level and no longer belong to a company.

The pre-migration script is required because the ORM only logs a warning and skips the constraint when the table already holds duplicates. It merges every group of records sharing a local code into the one owning an external ID, so the data of the module that declared it still resolves instead of being recreated as a new duplicate on its next update, then the active one, then the oldest. Everything referencing the discarded records is redirected to the kept one, which also recovers the descriptions only they had and is unarchived when any of them was still in use. Databases coming from 17.0 are the main target, as back then the same local code was allowed once per company.

Also removes security/product_hs_security.xml, left behind in the tree when it was dropped from the manifest: it is not loaded any more and its record rule filters on the removed company_id.

@Tecnativa
ping @sergio-teruel @carlos-lopez-tecnativa

### DIFF
--- a/product_harmonized_system/__manifest__.py
+++ b/product_harmonized_system/__manifest__.py
@@ -10,7 +10,7 @@
 
 {
     "name": "Product Harmonized System Codes",
-    "version": "18.0.2.0.0",
+    "version": "18.0.2.0.1",
     "category": "Reporting",
     "license": "AGPL-3",
     "summary": "Base module for Product Import/Export reports",
--- a/product_harmonized_system/migrations/18.0.2.0.1/pre-migration.py
+++ b/product_harmonized_system/migrations/18.0.2.0.1/pre-migration.py
@@ -1,0 +1,100 @@
+# Copyright 2026 Tecnativa - Carlos Dauden
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+import logging
+
+from openupgradelib import openupgrade
+from openupgradelib.openupgrade_merge_records import merge_records
+
+_logger = logging.getLogger(__name__)
+
+
+def _get_duplicated_codes(env):
+    """Return the groups of H.S. codes sharing the same local code.
+
+    Every group is a list of record ids sorted by the criteria electing the
+    record to keep, which therefore comes first:
+
+    1. The records owning an external ID, so the data of the module that
+       declared them still resolves. Otherwise that module would recreate the
+       record on its next update and reintroduce the duplicate.
+    2. The active ones, as the archived records are the discarded leftovers.
+    3. The oldest one, as it is the one the historical documents point to.
+    """
+    env.cr.execute(
+        """
+        WITH code AS (
+            SELECT
+                hc.id,
+                hc.local_code,
+                hc.active,
+                EXISTS (
+                    SELECT 1 FROM ir_model_data imd
+                    WHERE imd.model = 'hs.code' AND imd.res_id = hc.id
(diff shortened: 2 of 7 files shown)


carlosdauden:
@alexis-via @luc-demeyer are you agree with this changes?

luc-demeyer (changes requested):
Imho this PR doesn't fly (and also not the one that removed the company_id):

The EU requires more granular statistical data than the global system, one broad 6-digit HS category is frequently broken down into several distinct 8-digit Intrastat codes.
While the core 8-digit Intrastat (Combined Nomenclature/CN8) code is identical across all EU member states, individual European countries still maintain the legal right to add extra digits for their own national tracking, customs, or tax.

By dropping the company_id we lost this granularity.
In order to restore this broken granularity we may consider to move to a M2O relationship between the local_code and the hs_code whereby the local_code contains the extension digits per country.
Uniqueness check : hs_code+local_code should be unique per country.

Hence not an 15 minutes job to refactor the code in this way.

luc-demeyer:
An alternative to the M2O approach documented in my previous comment would be to add the country_id field to the HS code and enforce uniqueness on hs_code, country_id.
The database structure changes are minimal if we do this but we may end up with lots of duplicates for those countries that stick to the european 8 digit intrastat codes. 
By making the country_id a non-required field we can share these codes between those countries.

I am looking forward to your feedback on these suggestions.

carlosdauden:
Thanks for both suggestions, Luc. I agree that the current PR needs reconsideration before enforcing global uniqueness and merging existing records.

An optional country_id would minimize structural changes and allow countries to share common codes. However, it would not by itself solve classification selection for products shared across countries. Also, if you mean the existing hs_code field, it contains only six digits: uniqueness should apply to the full local_code within its scope, since several local codes can share the same six-digit HS code. Shared records would also need explicit uniqueness handling when country_id is NULL.

For the longer term, I would favor your first suggestion: separate the international six-digit HS code from the local classification codes, with multiple local codes referencing the same HS code. Each local code would contain its full code and its applicable country or shared nomenclature. This would allow countries using the same European nomenclature to share records without duplicating them.

We would also need to extend product classification. Products and categories currently hold a single hs_code_id, and the lookup does not consider the declaring country. I would propose country-specific classifications with a shared default where applicable, retaining the category fallback when the product has no classification.

When preparing a declaration, the lookup would select the classification applicable to the declaring country. We should not infer a national code from the six-digit HS code alone, since several national subdivisions can belong to it. If no applicable classification is configured, we should report it rather than silently select another country's code.

The migration should preserve existing assignments and only merge records known to represent the same classification within the same scope. The current migration, which merges solely by local_code, would therefore need to be revised once we agree on the model.

This is broader than the original fix, but it would address both uniqueness and classification selection. Would you agree with this direction?

luc-demeyer:
@carlosdauden 
If you have time to work on the 'long-term' solution, please go ahead.
If you want me to do it now, I would go for the simpler country_id based approach whereby we can categorize e.g. based upon the 6 char hs_code. We also have a lot less risk to break the localization modules with the simpler solution.

